PR #5, opened 2016-08-17
testEachStep_EVA-387

poggio84:
A new test for each Step of the pipeline - EVA-387


src/test/java/embl/ebi/variation/eva/pipeline/jobs/VariantStatsConfigurationTest.java
```diff
@@ -113,10 +107,9 @@ public void validLoadStats() throws JobExecutionException, IOException, IllegalA
         String input = VariantStatsConfigurationTest.class.getResource(FILE_20).getFile();
         VariantSource source = new VariantSource(input, "1", "1", "studyName");
         String dbName = VALID_LOAD_STATS;
-        String outputDir = input;
 
         pipelineOptions.put("input", input);
-        pipelineOptions.put("outputDir", outputDir);
+        pipelineOptions.put("outputDir", input);
```
cyenyxe: The input file and the output directory should probably not be the same?

cyenyxe:
Although the implemented tests are good, one of the main goals for this task was to replace the already existing tests, that require a whole job to be run every time, with truly independent step tests. Using JobLauncherTestUtils instead of manually configuring the jobs would allow to do this.

Unfortunately the "manual execution" used for the tests in this PR only allow the RepeatStatus to be checked, but not the ExitStatus that flags whether the step ran successfully.

PR #6, opened 2016-08-18
EVA-430: mapping from gene name to coordinates

jmmut:
This PR adds an extra step (and a standalone job) to set up a mongo collection `genes` (just like `variants` or `files`), listing all the genes with the region they comprise.

The input file is a GTF from a given species.


src/main/java/embl/ebi/variation/eva/pipeline/gene/GeneLineMapper.java
```diff
@@ -0,0 +1,49 @@
...
+            String[] keyValue = attribute.split(" ");
+
+            // don't do a `put(keyValue[0], keyValue[1])`: a space may appear before the key
+            String value = keyValue[keyValue.length - 1];
+            String valueWithoutQuotes = value.substring(1, value.length() - 1); // remove quotes
+            attributes.put(keyValue[keyValue.length - 2], valueWithoutQuotes);
+        }
+
+        return new GeneMongoBean(attributes.get("gene_id"), attributes.get("gene_name"), lineSplit[2], lineSplit[0],
+                Integer.parseInt(lineSplit[3]), Integer.parseInt(lineSplit[4]));
```
cyenyxe: Looking at how GTF represents transcripts, those lines should get "transcript_id" and "transcript_name" instead of "gene_id" and "gene_name". Then the GeneMongoBean class should be renamed to something like Feature or FeatureCoordinates (it could be serialized to other databases so the reference to Mongo can be dropped).

src/main/java/embl/ebi/variation/eva/pipeline/gene/GeneLineMapper.java
```diff
@@ -0,0 +1,49 @@
...
+        String[] lineSplit = line.split("\t");
+        String[] attributesSplit = lineSplit[8].split(";");
+        Map<String, String> attributes = new TreeMap<>();
+        for (String attribute : attributesSplit) {
+            String[] keyValue = attribute.split(" ");
+
+            // don't do a `put(keyValue[0], keyValue[1])`: a space may appear before the key
+            String value = keyValue[keyValue.length - 1];
+            String valueWithoutQuotes = value.substring(1, value.length() - 1); // remove quotes
+            attributes.put(keyValue[keyValue.length - 2], valueWithoutQuotes);
```
cyenyxe: Could use [String.replace(CharSequence target, CharSequence replacement)](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#replace-java.lang.CharSequence-java.lang.CharSequence-) instead.

jmmut: actually, I prefer not to use regex if I can, a quick profiling shows that this method takes 6% of the job's time, (0.8 seconds with a 500K lines GTF); and with the `replace` it rises to 12% of the job's time (2 seconds).

src/main/java/embl/ebi/variation/eva/pipeline/gene/GeneLineMapper.java
jmmut: fixed in next commit

PR #10, opened 2016-08-26
Refactoring of job, steps, readers and writers EVA-443

jorizci:
Refactored readers and writers to be independent non injectable stateless class.
Refactored the injection of tasklets and steps.
Code cleaning of common functions in jobs and steps.


src/main/java/embl/ebi/variation/eva/pipeline/jobs/CommonJobStepInitialization.java
```diff
@@ -0,0 +1,45 @@
...
+    private VariantJobsArgs variantJobsArgs;
+
+    @Autowired
+    private StepBuilderFactory stepBuilderFactory;
+
+    /**
+     * Initialize a Step with common configuration
+     * @param tasklet to be initialized with common configuration
+     */
+    protected void initStep(final TaskletStepBuilder tasklet) {
```
cyenyxe: This method seems to be used only from `generateStep` now. Could there be any benefit on keeping them separate?

jorizci: They can be merged, it's splitted to keep the original idea, and if we are
going to add more property initializations may be better if init is its own
function just for clarity sake.

2016-08-31 10:59 GMT+01:00 Cristina Yenyxe Gonzalez Garcia <
notifications@github.com>:

> ow. Could there be any benefit on keeping them separate?

src/main/java/embl/ebi/variation/eva/pipeline/steps/writers/VariantAnnotationWriter.java
```diff
@@ -0,0 +1,17 @@
+package embl.ebi.variation.eva.pipeline.steps.writers;
+
+import embl.ebi.variation.eva.pipeline.MongoDBHelper;
+import embl.ebi.variation.eva.pipeline.annotation.load.VariantAnnotationMongoItemWriter;
+import org.opencb.datastore.core.ObjectMap;
+
+/**
+ * TODO this class is the translation of the original code. This class can be removed lated if parent class can be refactored.
```
cyenyxe: Which refactoring would it require? I guess it mainly has to do with the method `writeVariantAnnotationInMongoDb`?

jorizci: A constructor in the parent class or initialisation code before using it...

src/main/java/embl/ebi/variation/eva/VariantJobsArgs.java
```diff
@@ -110,6 +117,7 @@
     private ObjectMap variantOptions  = new ObjectMap();
     private ObjectMap pipelineOptions  = new ObjectMap();
 
+    @PostConstruct
```
cyenyxe: Could this cause any conflicts when combined with invocations from test methods annotated with `@Before`? See for instance [the VariantsAnnotLoadTest class](https://github.com/EBIvariation/eva-pipeline/blob/develop/src/test/java/embl/ebi/variation/eva/pipeline/steps/VariantsAnnotLoadTest.java#L80).

jorizci: It should not.
The init method (`loadArgs`) can be called multiple times (for instance if you want to refresh the config properties) `@PostConstruct` definition makes the warranty that the method is executed after creation and before injection.

PR #13, opened 2016-09-05
EVA-447 Improved performance dumping variants to annotate

cyenyxe:
Performance has improved approximately 10 times by doing the following:
* Removing unnecessary sort removed
* Removing unnecessary projected field "type"
* Searching using index on "annot.ct.so" field

It is not amazing yet though (~150 variants/sec), probably due to the use of paginated instead of cursor-based queries, but is it a start.


jmmut:
This may break some tests as I remember one that relied on the variants being sorted. I guess it can be sorted afterwards for the test only

jmmut:
the problem and changes of this PR were already done in #16 and #20

PR #18, opened 2016-09-13
Added class documentation EVA-446

poggio84:
Documentation added

To be completed:
- GeneLoaded step
- DbInit step

EVA-446


src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VariantAnnotationReader.java
```diff
@@ -9,6 +24,17 @@
...
+ * Reader of VEP output annotation flat file. The file should be zipped.
+ *
+ * VEP output format:
+ *
+ *  20_60343_G/A	20:60343	A	-	-	-	intergenic_variant	-	-	-	-	-	-
+ *  20_60419_A/G	20:60419	G	-	-	-	intergenic_variant	-	-	-	-	-	-
+ *  20_60479_C/T	20:60479	T	-	-	-	intergenic_variant	-	-	-	-	-	rs149529999	GMAF=T:0.0018;AFR_MAF=T:0.01;AMR_MAF=T:0.0028
+ *  ...
+ *
+ */
```
poggio84: I'm moving this into the new VariantReader (cursor based) that is in develop

src/main/java/uk/ac/ebi/eva/pipeline/jobs/CommonJobStepInitialization.java
```diff
@@ -11,6 +26,11 @@
 
 import uk.ac.ebi.eva.pipeline.configuration.VariantJobsArgs;
 
+/**
+ * Helper class to build jobs.
+ *
+ * This is not intended to be used as an actual job, but to be extended in those actual job classes.
```
cyenyxe: If it's not intended to a job, shouldn't we make it abstract? @jorizci, is there any reason not to?

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/VariantsAnnotLoad.java
```diff
@@ -41,6 +41,13 @@
...
  * Step class that:
  * - READ: read a list of VEP {@link VariantAnnotation} from flat file
+ *
+ *  Example file content:
+ *  20_60343_G/A	20:60343	A	-	-	-	intergenic_variant	-	-	-	-	-	-
+ *  20_60419_A/G	20:60419	G	-	-	-	intergenic_variant	-	-	-	-	-	-
+ *  20_60479_C/T	20:60479	T	-	-	-	intergenic_variant	-	-	-	-	-	rs149529999	GMAF=T:0.0018;AFR_MAF=T:0.01;AMR_MAF=T:0.0028
+ *
+ *
  * - LOAD: write the {@link VariantAnnotation} into Mongo db
```
cyenyxe: Other classes are described in terms of "input" and "output". Could this one follow the same style?

src/main/java/uk/ac/ebi/eva/utils/MongoDBHelper.java
```diff
@@ -31,7 +31,7 @@
 /**
  * @author Diego Poggioli
  *
- *
+ * Utility class dealing with MongoDB connections using pipelineOptions
```
cyenyxe: "pipeline options" instead of the variable name would allow to provide those options in other ways without making the documentation look weird.

jmmut:
> Comments [...] should be associated to a method instead of a whole class

Those comments were intended to show the general requirements of the class tested, as general context. Not the actual test method logic, because the same requirement can be tested in several ways, e.g. for VariantAnnotationReader, "incorrect input lines should not make the reader fail." there are two tests. The alternative would be to repeat that comment on both tests. 

Agree that it's kind of prone to obsolescence, though.

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/VariantsAnnotLoad.java
jmmut: mmh, it's a different description. the input and output are black box descriptions, while these are explaining the internal workings. Will add some more documentation.

src/main/java/uk/ac/ebi/eva/utils/MongoDBHelper.java
jmmut: fixed in next commit

src/main/java/uk/ac/ebi/eva/pipeline/jobs/CommonJobStepInitialization.java
jmmut: talked with @jorizci, fixed in next commit

jorizci: It should be abstract, spring doesn't have problems with that, was probably a slip from my part.

PR #19, opened 2016-09-15
optionalAnnotationSteps_EVA-444

poggio84:
- New decider implemented to check that there are variants to annotate before running VEP. If the VEP input file is empty then VEP and annotations load will not run.
- Removed extension .gz to the vep input file

EVA-444


src/main/java/uk/ac/ebi/eva/pipeline/io/GzipLazyResource.java
```diff
@@ -49,5 +49,10 @@ public GzipLazyResource(String path) {
     public InputStream getInputStream() throws IOException {
         return new GZIPInputStream(super.getInputStream());
     }
+
+    public boolean isEmpty() throws IOException {
+        return getInputStream().read() == -1;
+    }
```
poggio84: No not used.. I will remove it

PR #20, opened 2016-09-19
bugFix: multiple run of VariantAnnotationMongoItemWriter overwrite annotations

poggio84:
uk.ac.ebi.eva.pipeline.io.writers.VariantAnnotationMongoItemWriter#doWrite has been updated using now the mongo $addToSet option. In this way annotations from previous batches are not overwritten by the new one. Using $addToSet to avoid duplicates and $each to append all the annotations for each id. 
Tests updated accordingly

EVA-460


jmmut:
This approach doesn't overwrite existing annotations. So if we ever need to overwrite, we will have to create an extra step for deleting annotations in all variants.

Having that in mind, I'm OK with this fix.

PR #30, opened 2016-10-20
EVA-507 add VariantOptionsConfigurerListener to ensure samples are written

jmmut:
Before, we used a `@PostConstruct` to set some job-dependent configuration into VariantOptions, but it seems that all jobs are instantiated even if they are not used, so the configuration was overwritten.

affected configuration:
- includeSamples
- compressGenotypes
- calculateStats
- includeStats

Now, we use a job listener to set up those in a "beforeJobExecution" listener.

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/listeners/VariantOptionsConfigurerListener.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/listeners/VariantOptionsConfigurerListener.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2016 EMBL - European Bioinformatics Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/listeners/VariantOptionsConfigurerListener.java
```diff
@@ -0,0 +1,61 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.batch.core.JobExecution;
+import org.springframework.batch.core.JobExecutionListener;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Import;
+import uk.ac.ebi.eva.pipeline.configuration.JobOptions;
+
+/**
+ * @author Jose Miguel Mut Lopez &lt;jmmut@ebi.ac.uk&gt;
```
cyenyxe: Class documentation.

jmmut: fixed in next commit

PR #31, opened 2016-10-21
EVA-501 using FeatureCoordinates from variation commons

jmmut:


pom.xml
```diff
@@ -21,8 +21,13 @@
         <biodata.version>0.4.5</biodata.version>
         <cellbase.version>3.1.3</cellbase.version>
     </properties>
-    
+
     <dependencies>
+        <dependency>
+            <groupId>uk.ac.ebi.eva</groupId>
+            <artifactId>variation-commons</artifactId>
+            <version>0.1</version>
```
sundarvenkata-EBI, line 29: This version should be changed to use the variation commons version after EVA-1984.

PR #34, opened 2016-11-03
EVA-23 Add a VCF reader (composable within spring steps)

jmmut:


src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VcfLineMapper.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @author Jose Miguel Mut Lopez &lt;jmmut@ebi.ac.uk&gt;
+ */
+public class VcfLineMapper implements LineMapper<List<Variant>> {
+    private VariantSource source;
+    private VariantVcfFactory factory;
+
+    public VcfLineMapper(VariantSource source) {
+        this.source = source;
+        factory = new VariantVcfFactory();
```
jmmut: To avoid a AggregatedVcfLineMapper and AggregatedVcfReader, here we could use the aggregation information in the `VariantSource` to decide which actual factory to use.

src/test/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReaderTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+        int count = 0;
+
+        while ((variants = vcfReader.read()) != null) {
+            count++;
+        }
+
+        // VcfReader should get all the lines from the file
+        long actualCount = JobTestUtils.getLines(new GZIPInputStream(new FileInputStream(inputFile)));
+        assertEquals(actualCount, count);
```
cyenyxe: The first argument should be expected value.

src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VcfLineMapper.java
cyenyxe: Actually I think it would be good to have them split, even if the class internally used is the same. Passing an aggregation mode "NONE" is a very implementation-specific detail, and it doesn't make much sense for a higher level layer.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReader.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * We need to fill the samples in the VariantSource, parsing the header before calling the factory
+     */
+    @Override
+    protected void doOpen() throws Exception {
+        VariantVcfReader reader = new VariantVcfReader(source, file.getAbsolutePath());
+        reader.open();
+        reader.pre();
+        source.addMetadata("variantFileHeader", reader.getHeader());
```
cyenyxe: How is this header accessible later on? We also need to store the information it contains. There is also a bug in the version of the Biodata library we use that misses meta entries with the same key (FORMAT, INFO, etc). This would be a very good opportunity to solve this issue.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReader.java
```diff
@@ -0,0 +1,66 @@
...
+ * @author Jose Miguel Mut Lopez &lt;jmmut@ebi.ac.uk&gt;
+ */
+public class VcfReader extends FlatFileItemReader<List<Variant>> {
+    private final VariantSource source;
+    private final File file;
+
+    public VcfReader(VariantSource source, File file) {
+        this.source = source;
+        this.file = file;
+        Resource resource = new GzipLazyResource(file);
```
cyenyxe: Maybe the system should detect whether the file is compressed or not and create a resource accordingly.

src/test/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReaderTest.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * {@link VcfReader}
+ * input: a Vcf file
+ * output: a VariantAnnotation each time its `.read()` is called
+ *
+ * incorrect input lines should not make the reader fail.
+ *
+ * @author Jose Miguel Mut Lopez &lt;jmmut@ebi.ac.uk&gt;
+ */
+public class VcfReaderTest {
```
cyenyxe: Some tests missing:

* Incorrect VCF file
* Testing that compressed and non-compressed VCF files are readable

src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VcfLineMapper.java
cyenyxe: Yes, that's for sure.

src/test/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReaderTest.java
jmmut: fixed in next commit

src/test/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReaderTest.java
jmmut: At some point I will stop doing this ¬¬.
fixed

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReader.java
jmmut: fixed in next commit

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReader.java
```diff
@@ -0,0 +1,253 @@
...
+public class VcfReader extends FlatFileItemReader<List<Variant>> {
+    /**
+     * You can retrieve the header of the VCF like this: `source.getMetadata().get(VARIANT_FILE_HEADER_KEY)`,
+     * but why "variantFileHeader"? we are using the converter
+     * {@link org.opencb.opencga.storage.mongodb.variant.DBObjectToVariantSourceConverter}, which requires it, and will
+     * store it in mongo as "header".
+     */
+    public static final String VARIANT_FILE_HEADER_KEY = "variantFileHeader";
+
+    private final VariantSource source;
```
cyenyxe: What will happen if the step fails while processing a variant? This object will be lost, right?

This could be implemented in another reader for header entries only, that will stop reading once the first line without a # is reached. And the variant reader would mark as comments the lines prefixed with #.

jmmut: done in next commit

src/test/java/uk/ac/ebi/eva/test/utils/JobTestUtils.java
```diff
@@ -159,4 +160,19 @@ public static File createTempFile() throws IOException {
     public static DBObject constructDbo(String variant) {
         return (DBObject) JSON.parse(variant);
     }
+
+    public static void uncompress(String inputCompressedFile, File outputFile) throws IOException {
+
```
jmmut: fixed in next commit

src/test/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReaderTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+        VcfReader vcfReader = new VcfReader(source, tempFile);
+        vcfReader.setSaveState(false);
+        vcfReader.open(executionContext);
+
+        consumeReader(inputFile, source, vcfReader);
+
+        char CHARACTER_TO_REPLACE_DOTS = (char) 163;
+        Map<String, Object> meta = source.getMetadata();
+        BasicDBObject metadataMongo = new BasicDBObject();
```
jmmut: We can move the method when we have a FileMongoWriter.

src/test/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReaderTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * This test is intended to check that the metadata of the VariantSource is properly filled when using a VcfReader,
+     * in a way that mostly maintains the structure we already have in mongo, in the files collection. See the doc in
+     * {@link VcfReader#prepareVariantSource()}
+     * @throws Exception
+     */
+    @Test
+    public void testConversion() throws Exception {
+
```
jmmut: fixed in next commit

src/test/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReaderTest.java
```diff
@@ -0,0 +1,205 @@
...
+        // consume the reader and check that a wrong variant raise an exception
+        exception.expect(FlatFileParseException.class);
+        while (vcfReader.read() != null) {
+        }
+    }
+
+    @Test
+    public void testUncompressedVcf() throws Exception {
+
```
jmmut: fixed in next commit

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReader.java
```diff
@@ -0,0 +1,253 @@
...
+            case A:
+                number = VCFConstants.PER_ALTERNATE_COUNT;
+                break;
+            case R:
+                number = VCFConstants.PER_ALLELE_COUNT;
+                break;
+            case G:
+                number = VCFConstants.PER_GENOTYPE_COUNT;
+                break;
+            case UNBOUNDED:
```
jmmut: yes, https://samtools.github.io/htsjdk/javadoc/htsjdk/constant-values.html#htsjdk.variant.vcf.VCFConstants.UNBOUNDED_ENCODING_v4. The API returns -1 for unknown number, but that constant is what's used for that purpose: [docs](https://samtools.github.io/htsjdk/javadoc/htsjdk/htsjdk/variant/vcf/VCFCompoundHeaderLine.html#getCount-htsjdk.variant.variantcontext.VariantContext-), [implementation](https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/vcf/VCFCompoundHeaderLine.java#L85)

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReader.java
```diff
@@ -0,0 +1,253 @@
...
+        setLineMapper(new VcfLineMapper(source));
+    }
+
+    public VcfReader(VariantSource source, String file) throws IOException {
+        this(source, new File(file));
+    }
+
+    @Override
+    protected void doOpen() throws Exception {
+        prepareVariantSource();
```
jmmut: well, it drops it [here](https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/vcf/VCFHeader.java#L115) an hardcodes it [here](https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/vcf/VCFHeader.java#L406) which is called if we retrieve the rows in the header.

As alternative, the [toString](https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/vcf/VCFHeader.java#L406) method  doesn't put the version, but we would need to reparse it to remove the tabs at the beginning of the line and the extra strings, and prepend the version which should be read by us.

conclusion, if it's a hard restriction, it's a serious argument against using htsjdk.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReader.java
```diff
@@ -0,0 +1,253 @@
...
+
+    public VcfReader(VariantSource source, File file) throws IOException {
+        this.source = source;
+        this.file = file;
+        Resource resource;
+        if (CompressionHelper.isGzip(file)) {
+            resource = new GzipLazyResource(file);
+        } else {
+            resource = new FileSystemResource(file);
+        }
```
jmmut: fixed in next commit

src/test/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReaderTest.java
```diff
@@ -0,0 +1,205 @@
...
+                                                 VariantStudy.StudyType.COLLECTION, VariantSource.Aggregation.NONE);
+
+        VcfReader vcfReader = new VcfReader(source, inputFile);
+        vcfReader.setSaveState(false);
+        vcfReader.open(executionContext);
+
+        // consume the reader and check that a wrong variant raise an exception
+        exception.expect(FlatFileParseException.class);
+        while (vcfReader.read() != null) {
+        }
```
jmmut: I would let this as is to enforce the rule of putting always the braces in `while`, `if` and `else` blocks

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReader.java
jmmut: fixed that bug in biodata in https://github.com/EBIvariation/biodata/pull/1, which is now a prerequisite for this PR

src/test/java/uk/ac/ebi/eva/pipeline/io/readers/VcfReaderTest.java
```diff
@@ -0,0 +1,205 @@
...
+        String studyName = "study name";
+        VariantSource source = new VariantSource(inputFile, fileId, studyId, studyName,
+                                                 VariantStudy.StudyType.COLLECTION, VariantSource.Aggregation.NONE);
+
+        VcfReader vcfReader = new VcfReader(source, inputFile);
+        vcfReader.setSaveState(false);
+        vcfReader.open(executionContext);
+
+        // consume the reader and check that a wrong variant raise an exception
+        exception.expect(FlatFileParseException.class);
```
jorizci: It can only be raised in the doRead method of FlatItemReader.

PR #35, opened 2016-11-07
MongoDB Variant writer

poggio84:
EVA-514

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriter.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriter.java
@@ -19,101 +19,70 @@ import com.mongodb.BasicDBObject;
 import com.mongodb.BulkWriteOperation;
 import com.mongodb.DBObject;
 import org.opencb.biodata.models.variant.Variant;
-import org.opencb.biodata.models.variant.VariantSourceEntry;
 import org.opencb.opencga.storage.mongodb.variant.DBObjectToVariantConverter;
-import org.opencb.opencga.storage.mongodb.variant.DBObjectToVariantSourceEntryConverter;
-import org.opencb.opencga.storage.mongodb.variant.DBObjectToVariantStatsConverter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.batch.item.data.MongoItemWriter;
 import org.springframework.data.mongodb.core.MongoOperations;
 import org.springframework.util.Assert;
+import uk.ac.ebi.eva.pipeline.model.converters.data.VariantToMongoDbObjectConverter;
+import uk.ac.ebi.eva.utils.MongoDBHelper;
 
 import java.util.List;
 
 /**
  * @author Diego Poggioli
- *         <p>
- *         Write a list of {@link Variant} into MongoDB
- *         See also {@link org.opencb.opencga.storage.mongodb.variant.VariantMongoDBWriter}
+ * <p>
+ * Write a list of {@link Variant} into MongoDB
+ * See also {@link org.opencb.opencga.storage.mongodb.variant.VariantMongoDBWriter}
  */
 public class VariantMongoWriter extends MongoItemWriter<Variant> {
     private static final Logger logger = LoggerFactory.getLogger(VariantMongoWriter.class);
 
-    private boolean includeStats;
     private MongoOperations mongoOperations;
     private String collection;
 
     private BulkWriteOperation bulk;
     private int currentBulkSize = 0;
 
-    private DBObjectToVariantConverter variantConverter;
-    private DBObjectToVariantStatsConverter statsConverter;
-    private DBObjectToVariantSourceEntryConverter sourceEntryConverter;
+    private VariantToMongoDbObjectConverter variantToMongoDbObjectConverter;
 
-    public VariantMongoWriter(boolean includeStats, String collection,
-                              DBObjectToVariantConverter variantConverter, DBObjectToVariantStatsConverter statsConverter,
-                              DBObjectToVariantSourceEntryConverter sourceEntryConverter, MongoOperations mongoOperations) {
-        this.includeStats = includeStats;
+    public VariantMongoWriter(String collection, MongoOperations mongoOperations,
+                              VariantToMongoDbObjectConverter variantToMongoDbObjectConverter) {
         this.collection = collection;
-        this.variantConverter = variantConverter;
-        this.statsConverter = statsConverter;
-        this.sourceEntryConverter = sourceEntryConverter;
         this.mongoOperations = mongoOperations;
+        this.variantToMongoDbObjectConverter = variantToMongoDbObjectConverter;
 
-        resetBulk();
+        bulk = mongoOperations.getCollection(collection).initializeUnorderedBulkOperation();
     }
 
     @Override
     protected void doWrite(List<? extends Variant> variants) {
         for (Variant variant : variants) {
-            variant.setAnnotation(null);
-            String id = variantConverter.buildStorageId(variant);
 
-            VariantSourceEntry variantSourceEntry = variant.getSourceEntries().entrySet().iterator().next().getValue();
+            String id = MongoDBHelper.buildStorageId(variant.getChromosome(), variant.getStart(),
+                    variant.getReference(), variant.getAlternate());
 
             // the chromosome and start appear just as shard keys, in an unsharded cluster they wouldn't be needed
             BasicDBObject query = new BasicDBObject("_id", id)
                     .append(DBObjectToVariantConverter.CHROMOSOME_FIELD, variant.getChromosome())
                     .append(DBObjectToVariantConverter.START_FIELD, variant.getStart());
 
-            BasicDBObject addToSet = new BasicDBObject()
-                    .append(DBObjectToVariantConverter.FILES_FIELD,
-                            sourceEntryConverter.convertToStorageType(variantSourceEntry));
-
-            if (includeStats) {
-                List<DBObject> sourceEntryStats = statsConverter.convertCohortsToStorageType(variantSourceEntry.getCohortStats(),
-                        variantSourceEntry.getStudyId(), variantSourceEntry.getFileId());
-                addToSet.put(DBObjectToVariantConverter.STATS_FIELD, new BasicDBObject("$each", sourceEntryStats));
-            }
-
-            if (variant.getIds() != null && !variant.getIds().isEmpty()) {
-                addToSet.put(DBObjectToVariantConverter.IDS_FIELD, new BasicDBObject("$each", variant.getIds()));
-            }
-
-            BasicDBObject update = new BasicDBObject()
-                    .append("$addToSet", addToSet)
-                    .append("$setOnInsert", variantConverter.convertToStorageType(variant));    // assuming variantConverter.statsConverter == null
+            DBObject update = variantToMongoDbObjectConverter.convert(variant);
 
             bulk.find(query).upsert().updateOne(update);
 
             currentBulkSize++;
-
-            executeBulk();
         }
+
+        executeBulk();
     }
 
     private void executeBulk() {
         if (currentBulkSize != 0) {
             logger.debug("Execute bulk. BulkSize : " + currentBulkSize);
             bulk.execute();
-            resetBulk();
         }
-    }
-
-    private void resetBulk() {
-        bulk = mongoOperations.getCollection(collection).initializeUnorderedBulkOperation();
-        currentBulkSize = 0;
     }
 
     @Override
--- a/src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriter.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriter.java
@@ -42,7 +42,6 @@ public class VariantMongoWriter extends MongoItemWriter<Variant> {
 
     private boolean includeStats;
     private String fileId;
-    private int bulkSize = 0;
     private DBCollection variantsCollection;
 
     private BulkWriteOperation bulk;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriter.java
```diff
@@ -0,0 +1,140 @@
...
+    private long numVariantsWritten;
+    private int currentBulkSize = 0;
+
+    private DBObjectToVariantConverter variantConverter;
+    private DBObjectToVariantStatsConverter statsConverter;
+    private DBObjectToVariantSourceEntryConverter sourceEntryConverter;
+
+    public VariantMongoWriter(boolean includeStats, String fileId, int bulkSize,
+                              DBObjectToVariantConverter variantConverter, DBObjectToVariantStatsConverter statsConverter,
+                              DBObjectToVariantSourceEntryConverter sourceEntryConverter, DBCollection variantsCollection) {
```
cyenyxe: This is quite different from [how the `VepAnnotationMongoWriter` was implemented](https://github.com/EBIvariation/eva-pipeline/blob/develop/src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationMongoWriter.java).

A `MongoOperations` object would define how to connect to the database. It supports bulk operations so it shouldn't be a problem. A `Converter` would wrap the OpenCGA ones.

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationMongoWriter.java
```diff
@@ -60,25 +60,12 @@
 public class VepAnnotationMongoWriter extends MongoItemWriter<VariantAnnotation> {
```
cyenyxe: See comments on `VariantMongoWriter` about what seems to be the recommended way to implement this. The previous implementation actually seemed to follow them more than the current one.

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/AnnotationLoaderStep.java
```diff
@@ -17,6 +17,7 @@
 package uk.ac.ebi.eva.pipeline.jobs.steps;
 
 import org.opencb.biodata.models.variant.annotation.VariantAnnotation;
+import org.opencb.opencga.storage.mongodb.variant.DBObjectToVariantAnnotationConverter;
```
cyenyxe: This import is not necessary.

src/main/java/uk/ac/ebi/eva/pipeline/listeners/ProcessedItemListener.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    public void beforeChunk(ChunkContext context) {
+    }
+
+    @Override
+    public void afterChunk(ChunkContext context) {
+        int itemProcessed = context.getStepContext().getStepExecution().getReadCount();
+        if(itemProcessed % 1000 == 0) {
+            logger.debug("Items processed " + itemProcessed);
```
cyenyxe: There are not really the "processed" items because some could have been skipped. We should track how many have been read and how many skipped.

Also even if the OpenCGA design was to check every 1000, since we can configure the chunk size in Spring Batch it could be best to just let the listener display the progress after each chunk, otherwise we will miss some notifications because of the different granularity.

src/main/java/uk/ac/ebi/eva/pipeline/model/converters/data/VariantToMongoDbObjectConverter.java
```diff
@@ -0,0 +1,97 @@
...
+        this.includeStats = includeStats;
+        this.statsConverter = calculateStats ? new DBObjectToVariantStatsConverter() : null;
+        DBObjectToSamplesConverter sampleConverter =
+                includeSample ? new DBObjectToSamplesConverter(true, samplesPosition) : null;
+        this.sourceEntryConverter = new DBObjectToVariantSourceEntryConverter(includeSrc, sampleConverter);
+        this.variantConverter = new DBObjectToVariantConverter(null, null);
+    }
+
+
+    public VariantToMongoDbObjectConverter(DBObjectToVariantConverter variantConverter,
```
cyenyxe: This constructor doesn't seem to be used.

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriterTest.java
```diff
@@ -0,0 +1,93 @@
...
+import java.net.UnknownHostException;
+import java.util.*;
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.when;
+
+/**
+ * Testing {@link VariantMongoWriter}
+ */
+@RunWith(ParallelRunner.class)
```
cyenyxe: This will not be necessary in the long term, as it can be set in the Maven POM file.

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriterTest.java
```diff
@@ -0,0 +1,93 @@
...
+public class VariantMongoWriterTest {
+    private VariantMongoWriter variantMongoWriter;
+    private VariantToMongoDbObjectConverter variantToMongoDbObjectConverter =
+            Mockito.mock(VariantToMongoDbObjectConverter.class);
+    private ObjectMap objectMap = new ObjectMap();
+    private final String collectionName = "variants";
+
+    @Before
+    public void setUp() throws Exception {
+        objectMap.put("config.db.read-preference", "primary");
```
cyenyxe: Please rebase develop and use the argument constants from class `JobParametersNames`.

src/test/java/uk/ac/ebi/eva/pipeline/model/converters/data/VariantToMongoDbObjectConverterTest.java
```diff
@@ -0,0 +1,144 @@
...
+        assertTrue(String.format("%s_%s_%s_%s", chromosome, start, reference, alternate).equals(setOnInsert.get("_id")));
+        assertTrue(chromosome.equals(setOnInsert.get("chr")));
+        assertEquals(start, setOnInsert.get("start"));
+        assertEquals(end, setOnInsert.get("end"));
+        assertTrue(reference.equals(setOnInsert.get("ref")));
+        assertTrue(alternate.equals(setOnInsert.get("alt")));
+    }
+
+    @Test
+    public void includeStatsTrueShouldIncludeStatistics() {
```
cyenyxe: Can the opposite be tested too?

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriter.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private void executeBulk() {
+        if (currentBulkSize != 0) {
+            logger.debug("Execute bulk. BulkSize : " + currentBulkSize);
+            bulk.execute();
+        }
+    }
+
+    @Override
+    public void afterPropertiesSet() throws Exception {
```
poggio84: Moved in constructor

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/AnnotationLoaderStep.java
poggio84: Removed

src/main/java/uk/ac/ebi/eva/pipeline/model/converters/data/VariantToMongoDbObjectConverter.java
poggio84: Removed

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriterTest.java
poggio84: Removed

src/test/java/uk/ac/ebi/eva/pipeline/model/converters/data/VariantToMongoDbObjectConverterTest.java
poggio84: The opposite is tested in the previous test where include and calculate Stats are false and there is the check `assertNull(addToSet.get("st"))`. The same for ids. Should I isolate this in a different test?

src/main/java/uk/ac/ebi/eva/pipeline/listeners/ProcessedItemListener.java
poggio84: Now reporting the number of read, written and skipped items. Class renamed into ChunkStatsListener. ChunkStatsReporterListener is another name option.

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriter.java
```diff
@@ -0,0 +1,94 @@
...
+        this.mongoOperations = mongoOperations;
+        this.variantToMongoDbObjectConverter = variantToMongoDbObjectConverter;
+
+        bulk = mongoOperations.getCollection(collection).initializeUnorderedBulkOperation();
+    }
+
+    @Override
+    protected void doWrite(List<? extends Variant> variants) {
+        for (Variant variant : variants) {
+
```
poggio84: done

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+    @Before
+    public void setUp() throws Exception {
+        objectMap.put(JobParametersNames.CONFIG_DB_READPREFERENCE, "primary");
+    }
+
+    @Test
+    public void noVariantsNothingShouldBeWritten() throws UnknownHostException {
+        String dbName = "VariantMongoWriterNoVariant";
+
+        objectMap.put("db.name", dbName);
```
cyenyxe: This should also use `JobParametersNames`.

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriterTest.java
```diff
@@ -0,0 +1,91 @@
...
+    public void variantsShouldBeWrittenIntoMongoDb() throws UnknownHostException {
+        String dbName = "VariantsShouldBeWrittenIntoMongoDb";
+
+        Variant variant = Mockito.mock(Variant.class);
+        when(variant.getChromosome()).thenReturn("1").thenReturn("2").thenReturn("3");
+        when(variant.getStart()).thenReturn(1).thenReturn(2).thenReturn(3);
+        when(variant.getReference()).thenReturn("A").thenReturn("B");
+        when(variant.getAlternate()).thenReturn("B").thenReturn("C");
+
+        objectMap.put("db.name", dbName);
```
cyenyxe: This should also use `JobParametersNames`.

src/test/java/uk/ac/ebi/eva/pipeline/model/converters/data/VariantToMongoDbObjectConverterTest.java
cyenyxe: Yes, because it's not clear under which conditions is each field written or not. Setting the boolean values inside the test methods instead of having a default value in the attributes would also help.

src/main/java/uk/ac/ebi/eva/pipeline/listeners/ChunkStatsListener.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public void afterChunk(ChunkContext context) {
+        logger.debug("Chunk stats: Items read count {}, items write count {}, items skip count{}",
+                context.getStepContext().getStepExecution().getReadCount(),
+                context.getStepContext().getStepExecution().getWriteCount(),
+                context.getStepContext().getStepExecution().getReadSkipCount()+
+                        context.getStepContext().getStepExecution().getProcessSkipCount()+
+                        context.getStepContext().getStepExecution().getWriteSkipCount()
```
cyenyxe: The code style guidelines recommend splitting before the + sign.

src/main/java/uk/ac/ebi/eva/pipeline/listeners/ChunkStatsListener.java
```diff
@@ -0,0 +1,51 @@
...
+public class ChunkStatsListener implements ChunkListener {
+    private static final Logger logger = LoggerFactory.getLogger(ChunkStatsListener.class);
+
+    @Override
+    public void beforeChunk(ChunkContext context) {
+    }
+
+    @Override
+    public void afterChunk(ChunkContext context) {
+        logger.debug("Chunk stats: Items read count {}, items write count {}, items skip count{}",
```
cyenyxe: I think this should have info level, because it's the main way to report progress on the process.
What do you think @poggio84 @jmmut @jorizci ?

src/main/java/uk/ac/ebi/eva/pipeline/listeners/ChunkStatsListener.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+package uk.ac.ebi.eva.pipeline.listeners;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.batch.core.ChunkListener;
+import org.springframework.batch.core.scope.context.ChunkContext;
+
+/**
+ * Log the number of read, write and skip items for each chunk.
```
cyenyxe: Is this actually a "per chunk" value? Being retrieved from the StepExecution I would assume it will display how much progress the step has made. In which case I would suggest a name like `StepProgressListener`.

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriterTest.java
poggio84: done

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriterTest.java
poggio84: done

src/main/java/uk/ac/ebi/eva/pipeline/listeners/ChunkStatsListener.java
cyenyxe: We can always increase the size of the chunk. In fact, 10 items per chunk might actually cause a lot of thread contention since the amount of computation will be quite low...

src/main/java/uk/ac/ebi/eva/pipeline/listeners/ChunkStatsListener.java
poggio84: Right.. I'm going to import @jmmut settings. Fixed

src/main/java/uk/ac/ebi/eva/pipeline/listeners/ChunkStatsListener.java
poggio84: Yes agree... will rename the class

PR #36, opened 2016-11-07
Spring boot changed to 1.4.1-RELEASE

jorizci:
Spring boot version changed to 1.4.1-RELEASE


cyenyxe:
Can the changes that don't apply to the Spring Boot upgrade be moved into another PR?

Please check the Javadoc on the way, because it seems the formatter got a bit crazy about it.

src/test/java/uk/ac/ebi/eva/pipeline/io/mappers/AnnotationLineMapperTest.java
```diff
@@ -15,13 +15,12 @@
  */
 package uk.ac.ebi.eva.pipeline.io.mappers;
 
-import static org.junit.Assert.assertNotNull;
```
cyenyxe: We shall decide on a import code style.

pom.xml
```diff
@@ -12,54 +12,63 @@
...
-        <version>1.3.3.RELEASE</version>
+        <version>1.4.1.RELEASE</version>
     </parent>
 
     <properties>
         <java.version>1.8</java.version>
         <opencga.version>0.5.2</opencga.version>
         <biodata.version>0.4.5</biodata.version>
         <cellbase.version>3.1.3</cellbase.version>
+        <postgresql.version>9.1-901-1.jdbc4</postgresql.version>
```
jorizci: Maybe we can remove it, I didn't want to remove the postgres just in case somebody was really using it (And I think is not the case...)

cyenyxe: We need to keep the PostgreSQL dependency because we are using it for the job repository database, but we could let Spring manage the driver version.

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/VepInputGeneratorStep.java
```diff
@@ -33,24 +26,27 @@
...
- *
  * This step dumps a list of variants without annotations from mongo.
- *
+ * <p>
  * input: mongo collection with the variants.
  * output: file with the list of variants, in a format to feed VEP ({@link VepInputFlatFileWriter})
- *
- *
+ * <p>
+ * <p>
```
cyenyxe: Double \<p\> here.

PR #41, opened 2016-11-21
EVA - 539

jorizci:
Annotation writer is injected depending on active profile.

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/Application.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/Application.java
@@ -34,6 +34,7 @@ import org.springframework.boot.autoconfigure.SpringBootApplication;
 public class Application {
 
     private static final Logger log = LoggerFactory.getLogger(Application.class);
+    public static final String VARIANT_ANNOTATION_MONGO_PROFILE = "variant-annotation-mongo";
 
     public static void main(String[] args) throws Exception {
         SpringApplication.run(Application.class, args);
--- a/src/main/java/uk/ac/ebi/eva/pipeline/configuration/AnnotationLoaderConfiguration.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/configuration/AnnotationLoaderConfiguration.java
@@ -1,0 +1,30 @@
+package uk.ac.ebi.eva.pipeline.configuration;
+
+import org.opencb.biodata.models.variant.annotation.VariantAnnotation;
+import org.springframework.batch.item.ItemWriter;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+import org.springframework.context.annotation.Scope;
+import org.springframework.data.mongodb.core.MongoOperations;
+import uk.ac.ebi.eva.pipeline.Application;
+import uk.ac.ebi.eva.pipeline.io.writers.VepAnnotationMongoWriter;
+import uk.ac.ebi.eva.utils.MongoDBHelper;
+
+@Configuration
+public class AnnotationLoaderConfiguration {
+
+    @Autowired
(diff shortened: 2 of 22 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/Application.java
```diff
@@ -34,6 +34,7 @@
 public class Application {
 
     private static final Logger log = LoggerFactory.getLogger(Application.class);
+    public static final String VARIANT_ANNOTATION_MONGO_PROFILE = "variant-annotation-mongo";
```
cyenyxe: We could use a more generic MongoDB profile, which would allow to easily configure **all** the readers and writers in the pipeline when changing a back-end, as well as simplify the .properties definition.

On the other hand, we wouldn't be able to test reading from an existing Mongo and writing into another technology under testing. But this may be not a use case to really consider. What do you think @jorizci ?

src/main/java/uk/ac/ebi/eva/pipeline/configuration/AnnotationLoaderConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+import org.springframework.context.annotation.Scope;
+import org.springframework.data.mongodb.core.MongoOperations;
+import uk.ac.ebi.eva.pipeline.Application;
+import uk.ac.ebi.eva.pipeline.io.writers.VepAnnotationMongoWriter;
+import uk.ac.ebi.eva.utils.MongoDBHelper;
+
+@Configuration
+public class AnnotationLoaderConfiguration {
```
cyenyxe: It would be more clearly linked to step if named `AnnotationLoaderStepConfiguration`. Or is there a better grouping we could do? :thinking:

src/main/java/uk/ac/ebi/eva/pipeline/Application.java
jorizci: I don't know if it will be interesting at some point mixing databases or just because we need to transfer data, simply i thought it's not a constraint that we need to add explicitly now. Anyway i think profiles also let you enable secondary profiles like if you put profile mongo-all adss profiles mongoA and mongoB. I'll return to you when i find the related documentation.

src/main/java/uk/ac/ebi/eva/pipeline/configuration/AnnotationLoaderConfiguration.java
jorizci: `AnnotationLoaderStepConfiguration` is fine by me, if we add more datasource configuration later then we can see if there is a better arrangement, but now i can't see anything better.

src/main/java/uk/ac/ebi/eva/pipeline/Application.java
jorizci: Yeah, I didnt' dream it. http://docs.spring.io/autorepo/docs/spring-boot/current/reference/html/boot-features-profiles.html#boot-features-adding-active-profiles

src/test/java/uk/ac/ebi/eva/pipeline/jobs/steps/VariantLoaderStepTest.java
```diff
@@ -56,6 +57,7 @@
  * Test for {@link VariantLoaderStep}
  */
 @RunWith(SpringRunner.class)
+@ActiveProfiles("variant-annotation-mongo")
```
cyenyxe: This class has nothing to do with the annotation, is it still necessary to enable that sub-profile for this to work?

jorizci: Yeah... surprisingly it does.
VariantLoaderStep->Autowires AnnotationFlowOptional at AnnotationFlow -> Autowires AnnotationLoad.

PR #44, opened 2016-11-30
EVA-486 PopulationStatistics: add chr, start, ref, alt fields

jmmut:
This was necessary if we want to do better queries like going
to a specific region. this would we harder if we had only the
merged vid. {vid: {$lte: 20_7000}} would not only query the
positions before 7000, but all the previous chromosomes as well.


src/test/java/uk/ac/ebi/eva/pipeline/io/writers/StatisticsMongoWriterTest.java
```diff
@@ -101,9 +101,10 @@ public void shouldCreateIndexesInCollection() throws Exception {
 
         // check vid has an index
```
jmmut: forgot about this PR, fixed in next commit

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/StatisticsMongoWriterTest.java
```diff
@@ -99,11 +103,12 @@ public void shouldCreateIndexesInCollection() throws Exception {
...
-        // check vid has an index
+        // check there is an index in chr + start + ref + alt + sid + cid
         assertEquals("[{ \"v\" : 1 , \"key\" : { \"_id\" : 1} , \"name\" : \"_id_\" , \"ns\" : \"" + databaseName +
-                        ".populationStatistics\"}, { \"v\" : 1 , \"unique\" : true , \"key\" : { \"vid\" : 1 , \"sid\" : 1 , " +
-                        "\"cid\" : 1} , \"name\" : \"vscid\" , \"ns\" : \"" + databaseName + ".populationStatistics\"}]",
-                statsCollection.getIndexInfo().toString());
+                             ".populationStatistics\"}, { \"v\" : 1 , \"unique\" : true , \"key\" : " +
+                             "{ \"chr\" : 1 , \"start\" : 1 , \"ref\" : 1 , \"alt\" : 1 , \"sid\" : 1 , \"cid\" : 1} , " +
+                             "\"name\" : \"vscid\" , \"ns\" : \"" + databaseName + ".populationStatistics\"}]",
+                     statsCollection.getIndexInfo().toString());
```
cyenyxe: This test could compare the DBObjects extracted from `getIndexInfo()` using the `equals()` method.

PR #40, opened 2016-11-21
Job parameter validation for VEP input generation and VEP run steps

poggio84:
Job parameters validator:
- VEP input generation step EVA-534
- VEP run step EVA-535

PR #45, opened 2016-12-01
Job parameter validation for VEP input generation and VEP run steps

poggio84:
Job parameters validator:

- VEP input generation step EVA-534
- VEP run step EVA-535


cyenyxe:
Supersedes #40

PR #47, opened 2016-12-05
EVA-572 Spring Data MongoDB jobparameters

cyenyxe:
Using `spring.data.mongodb.*` parameters to provide database connection details to the pipeline. I couldn't find a way to select the read preference, but it is not very relevant to the pipeline anyway.

`PopulationStatisticsLoaderStep` refactored to remove dependencies on OpenCGA manager classes. This allows it to use said Spring Data parameters.

### DIFF
--- a/examples/application.properties
+++ b/examples/application.properties
@@ -24,9 +24,16 @@ config.restartability.allow=false
 
 
 # MONGO DATABASE
+spring.data.mongodb.host=
+spring.data.mongodb.authentication-database=
+spring.data.mongodb.username=
+spring.data.mongodb.password=
+spring.data.mongodb.database=
 config.db.read-preference=primary
 db.collections.features.name=features
+db.collections.files.name=files
 db.collections.stats.name=populationStatistics
+db.collections.variants.name=variants
 
 
 # LOGGING
--- a/examples/application.properties
+++ b/examples/application.properties
@@ -25,14 +25,17 @@ config.restartability.allow=false
 
 # MONGO DATABASE
 spring.data.mongodb.host=
+spring.data.mongodb.port=
 spring.data.mongodb.authentication-database=
 spring.data.mongodb.username=
 spring.data.mongodb.password=
 spring.data.mongodb.database=
 config.db.read-preference=primary
 db.collections.features.name=features
-db.collections.files.name=files
 db.collections.stats.name=populationStatistics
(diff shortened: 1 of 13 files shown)


examples/application.properties
```diff
@@ -24,9 +24,16 @@ config.restartability.allow=false
...
 
 # MONGO DATABASE
+spring.data.mongodb.host=
+spring.data.mongodb.authentication-database=
+spring.data.mongodb.username=
+spring.data.mongodb.password=
+spring.data.mongodb.database=
 config.db.read-preference=primary
 db.collections.features.name=features
+db.collections.files.name=files
```
jmmut: It's misleading to put `files.name` and `variants.name` here, as they are not used, they are overwritten with the opencga ones

jmmut: However, this will be used when we remove the transform and load from opencga, in EVA-22

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/PopulationStatisticsLoaderStep.java
```diff
@@ -83,27 +103,120 @@
...
+        // Store variant statistics in Mongo
+        while (parser.nextToken() != null) {
+            variantsNumber++;
+            statsBatch.add(parser.readValueAs(VariantStatsWrapper.class));
+
+            if (statsBatch.size() == batchSize) {
+                QueryResult<?> writeResult = variantDBAdaptor.updateStats(statsBatch, options);
+                writes += writeResult.getNumResults();
+                logger.info("stats loaded up to position {}:{}", 
+                		statsBatch.get(statsBatch.size()-1).getChromosome(),
```
jmmut: tabs used here instead o spaces

examples/application.properties
cyenyxe: Yes, it is a "proactive" change. Shall I add a comment to clarify when will it apply?

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/PopulationStatisticsLoaderStep.java
cyenyxe: Bloody Eclipse XD

examples/application.properties
jmmut: yes, I think it will be clearer

PR #48, opened 2016-12-05
VariantsIndexes_EVA-550

poggio84:
The unique index on { chr : 1, start : 1, end : 1, ref : 1, alt : 1 } not generated yet.
(To discuss if we need to move the index generation into a beforeRead listener)

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriter.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriter.java
@@ -42,6 +42,10 @@ import java.util.List;
 public class VariantMongoWriter extends MongoItemWriter<Variant> {
     private static final Logger logger = LoggerFactory.getLogger(VariantMongoWriter.class);
 
+    public static final String BACKGROUND_INDEX = "background";
+
+    public static final String ANNOTATION_FIELD = "annot";
+
     private final MongoOperations mongoOperations;
 
     private final String collection;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VariantMongoWriter.java
```diff
@@ -81,4 +87,25 @@ private void executeBulk(BulkWriteOperation bulk, int currentBulkSize) {
...
+                                                 DBObjectToVariantSourceEntryConverter.STUDYID_FIELD);
+        String filesFileIdField = String.format("%s.%s", DBObjectToVariantConverter.FILES_FIELD,
+                                                DBObjectToVariantSourceEntryConverter.FILEID_FIELD);
+        mongoOperations.getCollection(collection).createIndex(
+                new BasicDBObject(filesStudyIdField, 1).append(filesFileIdField, 1).append("background", true));
+        mongoOperations.getCollection(collection).createIndex(new BasicDBObject(
+                "annot.xrefs.id", 1).append("background", true));
+        mongoOperations.getCollection(collection).createIndex(new BasicDBObject("annot", 1).append("background", true));
+        
+        //// TODO: 05/12/2016 to create a unique index on { chr : 1, start : 1, end : 1, ref : 1, alt : 1 } double check with DBA
```
poggio84: Is this necessary? By default, MongoDB creates a unique index on the _id field during the creation of a collection. See [https://docs.mongodb.com/v3.2/core/index-unique/](url)

cyenyxe: It is not necessary, it was stated more to clarify that this comment no longer applied :)

PR #56, opened 2016-12-14
EVA-532 Job parameters validator - variant normalization+store step

jmmut:


src/test/java/uk/ac/ebi/eva/pipeline/parameters/validation/step/VariantLoaderStepParametersValidatorTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private Map<String, JobParameter> parameters;
+
+    private Map<String, JobParameter> optionalParameters;
+
+    @Before
+    public void setUp() throws IOException {
+        validator = new VariantLoaderStepParametersValidator();
+        final File dir = TestFileUtils.getResource("/parameters-validation/");
+        final File inputVcf = TestFileUtils.getResource("/parameters-validation/file.vcf.gz");
```
cyenyxe: Better to use `temporaryFolderRule` and `temporaryFolderRule.newFile()`, so we don't need to set the flags to true. It will also prevent interactions among concurrent tests, where some could set them to false and others to true. Finally, if all the validations are run on temporary folders then we don't need the `parameters-validation` folder.

Also applies to `VepAnnotationGeneratorStepParametersValidatorTest`.

src/test/java/uk/ac/ebi/eva/pipeline/parameters/validation/step/VepAnnotationGeneratorStepParametersValidatorTest.java
```diff
@@ -49,31 +51,63 @@ public void allJobParametersAreValid() throws JobParametersInvalidException, IOE
...
+
+    }
+
+    @Test
+    public void allJobParametersAreValid() throws JobParametersInvalidException, IOException {
+        validator.validate(new JobParameters(parameters));
+    }
+
+    @Test(expected = JobParametersInvalidException.class)
+    public void APP_VEP_PATHIsMissing() throws JobParametersInvalidException, IOException {
```
cyenyxe: Normalize the capitalization to look as in other classes, i.e. `appVethPathIsMissing`. Also applies to `VepInputGeneratorStepParametersValidatorTest`.

src/test/java/uk/ac/ebi/eva/pipeline/parameters/validation/step/VariantLoaderStepParametersValidatorTest.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Test
+    public void allJobParametersIncludingOptionalAreValid() throws JobParametersInvalidException, IOException {
+        parameters.putAll(optionalParameters);
+        validator.validate(new JobParameters(parameters));
+    }
+
+    @Test(expected = JobParametersInvalidException.class)
+    public void dbNameIsRequired() throws Exception {
```
cyenyxe: Normalize the use of required/missing at the end of the method name across all the validator tests classes.

src/test/java/uk/ac/ebi/eva/pipeline/parameters/validation/step/VepAnnotationGeneratorStepParametersValidatorTest.java
jmmut: I did this to see what did you guys think, as this way it was copied (at the same time for all the entries with multiple edition) from the parameter in the line below. For the regular capitalization I couldn't find another way that writing one by one. Also, in upper case it's the actual name of the constant.

but well, just a failed experiment

src/test/java/uk/ac/ebi/eva/pipeline/parameters/validation/step/VariantLoaderStepParametersValidatorTest.java
jmmut: I'll go with is Required. That's what the test is checking. If the test fails it means that the field is not actually required.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/step/VariantLoaderStepParametersValidator.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private CompositeJobParametersValidator compositeJobParametersValidator() {
+        final List<JobParametersValidator> jobParametersValidators = Arrays.asList(
+                new DbNameValidator(),
+                new DbCollectionsVariantsNameValidator(),
+                new DbCollectionsFilesNameValidator(),
+                new InputStudyIdValidator(),
+                new InputVcfIdValidator(),
+                new InputVcfValidator(),
+                new OptionalValidator(new InputVcfAggregationValidator(), JobParametersNames.INPUT_VCF_AGGREGATION),
```
jmmut: At the moment it is necessary for both. the VariantLoaderStep (which GenotypedJob and AggregatedJob share) currently creates a bean for the VcfReader (AggregatedVcfReader or  VcfReader depending on this parameter). In some previous version, it was the job who defined the reader, but with @jorizci we decided this design was a bit better.

I think that making the aggregation not mandatory is not trivial, and may involve one of:
- default NONE in JobOptions, required for any aggregated, even BASIC
- not required for NONE or BASIC, but the jobs add this parameter to the listener (which modifies the JobOptions so the step instantiates the correct one).
- stop sharing the VariantLoaderStep in both jobs (this would make me very sad)

jmmut: I just realised I should put InputVcfAggregationValidator as a non-optional validator

cyenyxe: In the long term, I think the ideal situation would be the 1st or 2nd from those you list. But we can leave it as mandatory for now if refactoring it is complicated.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/InputVcfAggregationValidator.java
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.batch.core.JobParameters;
+import org.springframework.batch.core.JobParametersInvalidException;
+import org.springframework.batch.core.JobParametersValidator;
+
+import uk.ac.ebi.eva.pipeline.parameters.JobParametersNames;
+
+public class InputVcfAggregationValidator implements JobParametersValidator {
+
+    @Override
+    public void validate(JobParameters parameters) throws JobParametersInvalidException {
```
jmmut: I just saw that I didn't add the tests  InputVcfAggregationValidatorTest and InputVcfValidatorTest

PR #59, opened 2016-12-21
Minor stats improvement

jmmut:
commits:
- fix bug: disabling the creation of (unused) popStats collection. There is a collection "populationStatistics" that we are not using yet, but the spring framework is preparing the indexes in that collection. 
- allow to tune the batch size for stats calculation. A big study (10K+ samples) was having memory troubles, so it is useful to use our `config.chunk.size` to also configure the batchsize in opencga's statistics calculation. if the memory usage is too high, setting a lower chunk size (such as 50) reduces the memory usage in the variants load step and the stats calculation step.

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/model/PopulationStatistics.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/model/PopulationStatistics.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 EMBL - European Bioinformatics Institute
+ * Copyright 2016-2017 EMBL - European Bioinformatics Institute
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/model/PopulationStatistics.java
```diff
@@ -15,17 +15,23 @@
...
+
 import java.util.Map;
 
 /**
- * setters have package visibility, the user should use the constructor.
+ * Statistics related to a set of samples for a given variant.
+ * <p>
+ * Setters have package visibility, the user should use the constructor to instantiate valid objects.
  */
+@Profile(Application.MONGO_EXPERIMENTAL_PROFILE)
```
cyenyxe: Shouldn't the profile be applied to the `StatisticsMongoWriter` class too (or instead)? What would happen if the writer is injected and the model is not?

jmmut: The problematic class is the PopulationStatistics when it's automatically injected. There are no problems when the classes are manually injected, which is how it is done currently. 

In the future, when we actually use the PopulationStatistics class and related ones, we just have to remove the experimental profile.

PR #65, opened 2017-01-16
EVA-520 Use JobParameters in variant loader step

jorizci:
-Use JobParameters in variant loader step
-Refactored VariantSource, now reader uses the minimun parameters required.

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VcfLineMapper.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VcfLineMapper.java
@@ -17,7 +17,6 @@ package uk.ac.ebi.eva.pipeline.io.mappers;
 
 import org.opencb.biodata.models.variant.VariantSource;
 import org.springframework.batch.item.file.LineMapper;
-
 import uk.ac.ebi.eva.commons.models.data.Variant;
 
 import java.util.List;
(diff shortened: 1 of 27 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VcfLineMapper.java
```diff
@@ -31,25 +30,21 @@
...
         this.factory = new VariantVcfFactory();
     }
 
     @Override
     public List<Variant> mapLine(String line, int lineNumber) {
         assertNotNull(this.getClass().getSimpleName() + " should be used to read genotyped VCFs only " +
-                              "(hint: set VariantSource.Aggregation to NONE)",
-                      factory);
-        return factory.create(source, line);
+                        "(hint: set VariantSource.Aggregation to NONE)", factory);
```
jmmut: This assert can not fail. It was here in the previous version because the constructor could throw an exception, leaving the `factory` uninitialized.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/steps/VariantLoaderStepTest.java
```diff
@@ -99,7 +100,13 @@ public void loaderStepShouldLoadAllVariants() throws Exception {
                 VariantSource.Aggregation.NONE));
```
jmmut: I would say this variant source is not needed anymore

jmmut: and the dbname is not needed either

jorizci: They are still on Job Options and Cristina disagreed to remove it completely until no one requires them.

PR #66, opened 2017-01-17
EVA-587 Fixed structure of "meta" field in "files" collection

cyenyxe:
With the change to a straightforward entity writing, the header dump in the "files" collection began to be written to "meta.variantFileHeader" instead of the correct "meta.header". This behavior has been reverted and a new test added.

The order of the `VariantSourceEntity` class attributes has been slightly modified to keep closer those that are related.

### DIFF
--- a/src/test/java/uk/ac/ebi/eva/pipeline/io/writers/VariantSourceEntityMongoWriterTest.java
+++ b/src/test/java/uk/ac/ebi/eva/pipeline/io/writers/VariantSourceEntityMongoWriterTest.java
@@ -86,19 +86,19 @@ public class VariantSourceEntityMongoWriterTest {
         while (cursor.hasNext()) {
             count++;
             DBObject next = cursor.next();
-            assertNotNull(next.get("fname"));
-            assertNotNull(next.get("fid"));
-            assertNotNull(next.get("sid"));
-            assertNotNull(next.get("sname"));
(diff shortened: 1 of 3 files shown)


src/test/java/uk/ac/ebi/eva/pipeline/io/writers/VariantSourceEntityMongoWriterTest.java
```diff
@@ -86,15 +86,23 @@ public void shouldWriteAllFieldsIntoMongoDb() throws Exception {
...
+            assertNotNull(next.get("sname"));
+            assertNotNull(next.get("samp"));
+            assertNotNull(next.get("stype"));
+            assertNotNull(next.get("date"));
+            assertNotNull(next.get("aggregation"));
+
+            DBObject meta = (DBObject) next.get("meta");
+            assertNotNull(meta);
+            assertNotNull(meta.get("fileformat"));
+            assertNotNull(meta.get("header"));
```
jmmut: shouldn't we use constants here? at least the ones easily available as this one. One might say that the tests would be too coupled, but I think it would show where do those strings come from. A disadvantage of the current way, is that it may suggest that you should handle this object with hardcoded strings.

On a side note, I wonder whether the field names in VariantSourceEntity as "fname" can be retrieved with reflection.

PR #68, opened 2017-01-18
EVA-587 Mapping dots in Mongo keys to another character

cyenyxe:
When using OpenCGA converters to store `VariantSource` objects, dots in samples names were mapped to another character (this is a Mongo requirement).

This behavior is now replicated using the class `MappingMongoConverter` from Spring Data. In order to provide instances of this class to build `MongoTemplate` objects, `MongoDBHelper` must support being autowired, which involves:

* It must be annotated as `@Configuration`
* Those methods creating `MongoTemplate` objects can't be static anymore
* Reader / writer / tasklet-step configurations need to `@Import({ MongoDBHelper.class })`, to ensure they will work both in tests and stand-alone runs

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
@@ -36,7 +36,7 @@ public class NonAnnotatedVariantsMongoReader extends MongoDbCursorItemReader {
                                            MongoConnection connection) throws UnknownHostException {
         super();
 
-        setMongo(MongoDBHelper.getMongoClient(connection));
+        setMongo(new MongoDBHelper().getMongoClient(connection));
 
         setDatabaseName(databaseName);
         setCollectionName(collectionsVariantsName);
--- a/src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 EMBL - European Bioinformatics Institute
+ * Copyright 2016-2017 EMBL - European Bioinformatics Institute
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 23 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
```diff
@@ -36,7 +36,7 @@ public NonAnnotatedVariantsMongoReader(String databaseName, String collectionsVa
                                            MongoConnection connection) throws UnknownHostException {
         super();
 
-        setMongo(MongoDBHelper.getMongoClient(connection));
+        setMongo(new MongoDBHelper().getMongoClient(connection));
```
jorizci: Don't like the idea of creating a new instance of a class that we are injecting... The method is a helper one... but the class now is a configuration...

Maybe we should move the method to static?

src/main/java/uk/ac/ebi/eva/utils/MongoDBHelper.java
```diff
@@ -16,42 +16,62 @@
 
 package uk.ac.ebi.eva.utils;
```
jorizci: Maybe this class should be moved.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
cyenyxe: We can't make it static because it requires the injection of the mapping converter, which is the reason for this refactoring (read the ticket description damn you! :stuck_out_tongue: ).

This was my mind slipping though, this class should look like [GeneWriterConfiguration](https://github.com/EBIvariation/eva-pipeline/pull/68/files/d2ba53ebe1fd80ec00afe3ca60a985fb72f03dbc#diff-d4abd9d30e525fc4f6584eaab081082e) does now.

src/main/java/uk/ac/ebi/eva/utils/MongoDBHelper.java
cyenyxe: Maybe split in two? The utility static methods could stay in the `utils` package, and the `getMongoOperations()` and related ones be moved to a new class `uk.ac.ebi.eva.pipeline.configuration.MongoConfiguration`.

jorizci: Yeah, it's a solution.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
jorizci: NonAnnotatedVariantsMongoReader should be harmonised with the rest of the readers/writers. The parent reader should be implemented again with the inclusion of template functionality.

src/main/java/uk/ac/ebi/eva/utils/MongoDBHelper.java
```diff
@@ -129,8 +47,7 @@ private MappingMongoConverter getMappingMongoConverter(MongoDbFactory mongoFacto
         return serverAddresses;
     }
 
-
-    private ReadPreference getMongoTemplateReadPreferences(String readPreference) {
+    public static ReadPreference getMongoTemplateReadPreferences(String readPreference) {
         switch (readPreference) {
             case "primary":
```
jmmut: a bit offtopic, but why isn't here primaryPrefered and secondaryPreferred?

cyenyxe: In fact, the `ReadPreference` enum from the Mongo API can deal with this perfectly, it even throws an `IllegalArgumentException` if an incorrect value is provided.

I will replace this method with a call to `ReadPreference.valueOf(readPreference)`.

PR #69, opened 2017-01-19
Use JobParameters in VEP run step - EVA-594

poggio84:
EVA-594

### DIFF
--- a/src/test/java/uk/ac/ebi/eva/pipeline/jobs/AnnotationJobTest.java
+++ b/src/test/java/uk/ac/ebi/eva/pipeline/jobs/AnnotationJobTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 EMBL - European Bioinformatics Institute
+ * Copyright 2016-2017 EMBL - European Bioinformatics Institute
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -28,6 +28,7 @@ import org.opencb.opencga.storage.mongodb.variant.DBObjectToVariantAnnotationCon
 import org.springframework.batch.core.BatchStatus;
 import org.springframework.batch.core.ExitStatus;
 import org.springframework.batch.core.JobExecution;
+import org.springframework.batch.core.JobParameters;
 import org.springframework.batch.core.StepExecution;
 import org.springframework.batch.test.JobLauncherTestUtils;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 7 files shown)


src/test/java/uk/ac/ebi/eva/pipeline/jobs/AnnotationJobTest.java
```diff
@@ -83,7 +87,19 @@
...
+        JobParameters jobParameters = new EvaJobParameterBuilder()
+                .vepPath(getResource(MOCK_VEP).getPath())
+                .vepCacheVersion("")
+                .vepCachePath("")
+                .vepCacheSpecies("")
+                .inputFasta("")
+                .vepNumForks("")
+                .outputDirAnnotation(OUTPUT_DIR_ANNOTATION)
+                .inputStudyId(INPUT_STUDY_ID)
+                .inputVcfId(INPUT_VCF_ID)
```
cyenyxe: It would be helpful for reading if these calls could be sorted alphabetically. Applies to the other tests classes.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/AnnotationJobTest.java
```diff
@@ -83,7 +87,19 @@
...
+                .toJobParameters();
+
+        JobExecution jobExecution = jobLauncherTestUtils.launchJob(jobParameters);
 
         assertEquals(ExitStatus.COMPLETED, jobExecution.getExitStatus());
         assertEquals(BatchStatus.COMPLETED, jobExecution.getStatus());
```
cyenyxe: Any ideas on how to make this test fail with so many empty parameters? The true VEP will break, but of course our mock doesn't :thinking:

Applies to the other test classes that depend on the VEP run step.

poggio84: Well in the near future all these parameters will be validated at the beginning so if we reach this point we are sure that they are valid... and not empty ;)

src/test/java/uk/ac/ebi/eva/pipeline/jobs/AnnotationJobTest.java
```diff
@@ -65,6 +66,9 @@
 public class AnnotationJobTest {
     private static final String MOCK_VEP = "/mockvep.pl";
     private static final String MONGO_DUMP = "/dump/VariantStatsConfigurationTest_vl";
+    private static final String OUTPUT_DIR_ANNOTATION = "/tmp/";
```
cyenyxe: Use a `PipelineTemporaryFolderRule` instead of a hardcoded path. Applies to other classes using the `outputDirAnnotation(...)` method.

poggio84: I think that this can only be done when all the annotation steps are using the jobParameters. If we mix the rule and properties then the files might end up in different folders. The rule for example is used in VepAnnotationGeneratorStepTest because is self-contained.

PR #70, opened 2017-01-23
EVA-608 Use JobParameters in VCF header loader step

jmmut:
The main change is in  src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/tasklets/FileLoaderStep.java and its test.

### DIFF
--- a/src/test/java/uk/ac/ebi/eva/utils/EvaJobParameterBuilder.java
+++ b/src/test/java/uk/ac/ebi/eva/utils/EvaJobParameterBuilder.java
@@ -17,45 +17,51 @@ package uk.ac.ebi.eva.utils;
 
 import org.springframework.batch.core.JobParameter;
 import org.springframework.batch.core.JobParametersBuilder;
+
 import uk.ac.ebi.eva.pipeline.parameters.JobParametersNames;
 
 import java.sql.Timestamp;
 import java.util.Date;
 
-public class EvaJobParameterBuilder extends JobParametersBuilder{
+public class EvaJobParameterBuilder extends JobParametersBuilder {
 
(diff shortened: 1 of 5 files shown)


src/test/java/uk/ac/ebi/eva/utils/EvaJobParameterBuilder.java
```diff
@@ -17,45 +17,51 @@
...
 import java.sql.Timestamp;
 import java.util.Date;
 
-public class EvaJobParameterBuilder extends JobParametersBuilder{
+public class EvaJobParameterBuilder extends JobParametersBuilder {
 
-    public EvaJobParameterBuilder inputStudyId(String inputStudyId){
-        addParameter(JobParametersNames.INPUT_STUDY_ID, new JobParameter(inputStudyId));
+    public EvaJobParameterBuilder inputStudyId(String inputStudyId) {
+        super.addParameter(JobParametersNames.INPUT_STUDY_ID, new JobParameter(inputStudyId));
```
jorizci: why?

jmmut: because I added a generic method at the end called addParameter as well that takes a string instead of the parent's one that takes a JobParameter.

cyenyxe: Then you could simply use `addString(JobParametersNames.INPUT_STUDY_ID, inputStudyId)` from the `JobParameterBuilder` class.

jmmut: oh, I didn't see that one. PR updated.

PR #73, opened 2017-01-25
EVA-594 VepGeneration_jobParameter_improvement2

poggio84:
Tasklet parameters retieved from autowired container and not from chunkContext

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/parameters/AnnotationParameters.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/parameters/AnnotationParameters.java
@@ -86,11 +86,11 @@ public class AnnotationParameters {
         return inputFasta;
     }
 
-    public String getVepInput(){
+    public String getVepInput() {
         return VepUtils.resolveVepInput(outputDirAnnotation, studyId, fileId);
     }
 
-    public String getVepOuput(){
+    public String getVepOuput() {
         return VepUtils.resolveVepOutput(outputDirAnnotation, studyId, fileId);
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/parameters/AnnotationParameters.java
```diff
@@ -0,0 +1,101 @@
...
+    public String getVepNumForks() {
+        return vepNumForks;
+    }
+
+    public String getInputFasta() {
+        return inputFasta;
+    }
+
+    public String getVepInput(){
+        return VepUtils.resolveVepInput(outputDirAnnotation, studyId, fileId);
```
cyenyxe: Do we need an extra class just for this now? It could be a private method.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/AnnotationParameters.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public String getVepOuput(){
```
cyenyxe: Whitespace missing between the bracket and the curly brace.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/AnnotationParameters.java
jmmut: It's fine for me both in a dedicated class or as a method VepAnnotationGeneratorStep. But if it's a method, at least make it package-private instead of private, as it is used in VepAnnotationGeneratorStep and VepAnnotationGeneratorStepTest

cyenyxe: I didn't notice it was used from other classes. In that case it may be better to wait until that step is refactored too.

PR #75, opened 2017-01-26
EVA-587 Unique index in "files" collection

cyenyxe:
A new unique, background index is now created in the "files" collection, on the following fields:

* File ID (`fid`)
* Filename (`fname`)
* Study ID (`sid`)

Both file ID and name because the file ID matches an ERZ, which can contain multiple files. We should consider renaming this field in the future.


src/test/java/uk/ac/ebi/eva/pipeline/io/writers/VariantSourceEntityMongoWriterTest.java
```diff
@@ -147,7 +150,34 @@ public void shouldWriteSamplesWithDotsInName() throws Exception {
...
+
+        VariantSourceEntity variantSourceEntity = getVariantSourceEntity();
+        filesWriter.write(Collections.singletonList(variantSourceEntity));
+
+        List<DBObject> indexInfo = fileCollection.getIndexInfo();
+
+        Set<String> createdIndexes = indexInfo.stream().map(index -> index.get("name").toString())
+                .collect(Collectors.toSet());
+        Set<String> expectedIndexes = new HashSet<>();
+        expectedIndexes.addAll(Arrays.asList("unique_file", "_id_"));
```
jorizci: I'd move the strings to constants.

PR #79, opened 2017-01-27
EVA-595 Use JobParameters in annotation loader step

jmmut:
Note:
- VariantAnnotationReaderConfiguration now takes a AnnotationParameters instead of JobOptions
- VariantAnnotationWriterConfiguration now takes a DatabaseParameters instead of JobOptions
- added EmptyVepInputDecider: the paramenter AnnotationDirectory
    should not be available through jobOptions, and the jobParameters
    is only available after instantiating the AnnotationFlow, so we
    need a way to resolve the vepInput path in a delayed way.
- removed jobOptions from AnnotationJobTest,
    AnnotationLoaderStepTest and VepInputGeneratorStepTest.


src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/EmptyVepInputDecider.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2016 EMBL - European Bioinformatics Institute
```
cyenyxe: Licensing year should be 2017.

src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/EmptyVepInputDecider.java
```diff
@@ -0,0 +1,74 @@
...
+
+import uk.ac.ebi.eva.pipeline.parameters.JobParametersNames;
+import uk.ac.ebi.eva.utils.VepUtils;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+/**
+ * Decider used to skip step(s) in case a previous step is generating an empty file
```
cyenyxe: Should state that this works not any empty file but the VEP input specifically.

src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/EmptyVepInputDecider.java
```diff
@@ -0,0 +1,74 @@
...
+            return new FlowExecutionStatus(STOP_FLOW);
+        }
+
+        return new FlowExecutionStatus(CONTINUE_FLOW);
+    }
+
+    private String getVepInput(JobExecution jobExecution) {
+        JobParameters jobParameters = jobExecution.getJobParameters();
+
+        return VepUtils.resolveVepInput(
```
cyenyxe: I thought the methods from `VepUtils` were to be moved to `URLHelper`, to have all the path handling placed together?

src/test/java/uk/ac/ebi/eva/pipeline/jobs/steps/AnnotationLoaderStepTest.java
```diff
@@ -73,21 +71,35 @@
...
+        String collectionVariantsName = "variants";
+        String studyId = "1";
+        String fileId = "1";
+        String vepOutput = VepUtils.resolveVepOutput(annotationFolder, studyId, fileId);
+        String vepOutputName = Paths.get(vepOutput).getFileName().toString();
+        temporaryFolderRule.newGzipFile(VepOutputContent.vepOutputContent, vepOutputName);
+
+        mongoRule.restoreDump(getResourceUrl(MONGO_DUMP), dbName);
+
+        JobParameters jobParameters = new EvaJobParameterBuilder()
```
cyenyxe: The input study ID and the input VCF ID shouldn't be necessary for the annotation loader, see [AnnotationLoaderStepParametersValidator](https://github.com/EBIvariation/eva-pipeline/blob/develop/src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/step/AnnotationLoaderStepParametersValidator.java). They are for the VEP input generation step.

jmmut: they both are used to resolve the filename of vepOutput

cyenyxe: Please then fix the step validator, because they are not required there.

PR #81, opened 2017-02-01
EVA-518 Use temporary folders in tests

jmmut:
related things done:
- MongoConnection is now a service. can be injected directly in tests (with `@TestPropertySource("classpath:test-mongo.properties")`) and `@ContextConfiguration(BaseTestConfiguration.class)`.
- Some non-temporal directories were retrieved from jobOptions. I removed other usages of jobOptions like obtaining the collection names (now they are just class constants).


src/test/java/uk/ac/ebi/eva/pipeline/jobs/AggregatedVcfJobTest.java
```diff
@@ -70,42 +66,26 @@
...
-
-    private static String opencgaHome = System.getenv("OPENCGA_HOME") != null ? System
-            .getenv("OPENCGA_HOME") : "/opt/opencga";
-
     public static final Set<String> EXPECTED_REQUIRED_STEP_NAMES = new TreeSet<>(
             Arrays.asList(BeanNames.LOAD_VARIANTS_STEP, BeanNames.LOAD_FILE_STEP));
 
     @Test
     public void aggregatedTransformAndLoadShouldBeExecuted() throws Exception {
-        Config.setOpenCGAHome(opencgaHome);
```
jorizci: Don't we need to set OpenCGA anymore?

jmmut: mm, yes we need it, I'll revert this

src/main/java/uk/ac/ebi/eva/pipeline/parameters/InputParameters.java
```diff
@@ -63,7 +63,6 @@
     @Value(PARAMETER + JobParametersNames.CONFIG_CHUNK_SIZE + "']?:1000}")
     private int chunkSize;
```
cyenyxe: I agree with the comment that the last 3 parameters should be moved to a different class. Another PR?

src/main/java/uk/ac/ebi/eva/pipeline/parameters/MongoConnection.java
```diff
@@ -0,0 +1,48 @@
+package uk.ac.ebi.eva.pipeline.parameters;
+
+import com.mongodb.ReadPreference;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+
+@Service
+public class MongoConnection {
```
cyenyxe: Please add class documentation and license declaration.

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/VariantSourceEntityMongoWriterTest.java
```diff
@@ -148,15 +152,10 @@ public void shouldWriteSamplesWithDotsInName() throws Exception {
...
-                VariantStorageManager.VARIANT_SOURCE);
-        String fileId = source.getFileId();
-        String studyId = source.getStudyId();
-        String studyName = source.getStudyName();
-        VariantStudy.StudyType studyType = source.getType();
-        VariantSource.Aggregation aggregation = source.getAggregation();
-
-        VcfHeaderReader headerReader = new VcfHeaderReader(new File(input), fileId, studyId, studyName,
+        VariantStudy.StudyType studyType = VariantStudy.StudyType.COLLECTION;
+        VariantSource.Aggregation aggregation = VariantSource.Aggregation.NONE;
```
cyenyxe: The rest of fields are class attributes.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/GenotypedVcfJobTest.java
```diff
@@ -88,45 +83,41 @@
...
+        File inputFile = getResource(INPUT_FILE);
+        String dbName = mongoRule.getRandomTemporaryDatabaseName();
+
+        String outputDirStats = temporaryFolderRule.newFolder().getAbsolutePath();
+        String outputDirAnnotation = temporaryFolderRule.newFolder().getAbsolutePath();
+
+        File variantsStatsFile = new File(URLHelper.getVariantsStatsUri(outputDirStats, INPUT_STUDY_ID, INPUT_VCF_ID));
+        File sourceStatsFile = new File(URLHelper.getSourceStatsUri(outputDirStats, INPUT_STUDY_ID, INPUT_VCF_ID));
+        assertFalse(variantsStatsFile.exists());
+        assertFalse(sourceStatsFile.exists());
```
cyenyxe: These asserts and the ones below should not be necessary because we are using temporary folders.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/InputParameters.java
jmmut: Yes, it can be done in another pull request. I'm not sure about some details, like should we leave them as job parameter injection ("#{jobParameters['config.chunk.size']}") or direct injection ("${config.chunk.size}"). Maybe we would want to change the chunk size depending on the job, but it only will affect the populationStatistics step, because the other ones are built before the jobparameters are resolved, so a change in a job parameter won't affect the job instantiation

jorizci: Related to this issue: https://www.ebi.ac.uk/ena/jira/browse/EVA-609

PR #82, opened 2017-02-01
EVA-636 Use JobParameters in all the steps not currently within any job

jmmut:
Instead of making PedLoaderStep use JobParameters, it should be a reader, so I changed PedLoaderStep to PedReader and made the tests uniform.

Also, some more tests were improved, removing unneeded jobOptions
and extracting constants


src/main/java/uk/ac/ebi/eva/pipeline/io/readers/PedReader.java
```diff
@@ -0,0 +1,54 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(PedReader.class);
+
+    private String pedigreePath;
+
+    public PedReader(String pedigreePath) {
+        this.pedigreePath = pedigreePath;
+    }
+
+    /**
+     * Maybe we should return null after the first call? this reader is intended to read just one element.
```
cyenyxe: Yes, the reading loop will never stop unless the method returns null at some point.

src/test/java/uk/ac/ebi/eva/pipeline/io/readers/PedReaderTest.java
```diff
@@ -0,0 +1,73 @@
...
+ * input: a pedigree file
+ * output: a Pedigree when method `.read()` is called.
+ */
+public class PedReaderTest {
+    private static final String PEDIGREE_FILE = "/ped/pedigree-test-file.ped";
+
+    private static final String MALFORMED_PEDIGREE = "/ped/malformed-pedigree-test-file.ped";
+
+    @Test
+    public void allPedFileShouldBeParsedIntoPedigree() throws Exception {
```
cyenyxe: Should be `wholePedFile` instead of `allPedFile`.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/AnnotationJobTest.java
```diff
@@ -137,7 +138,7 @@ public void allAnnotationStepsShouldBeExecuted() throws Exception {
         int consequenceTypeCount = 0;
         while (cursor.hasNext()) {
             count++;
-            DBObject dbObject = (DBObject) cursor.next().get("annot");
+            DBObject dbObject = (DBObject) cursor.next().get(VariantToDBObjectConverter.ANNOTATION_FIELD);
```
cyenyxe: This is statically imported in some classes and non-statically in others. I don't have a particular preference for any of those forms, but it should be homogeneous.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/PedReader.java
jmmut: I left it like this because it is very similar to the VcfHeaderReader, which is doing the same (parsing a file), and we actually use the VcfHeaderReader from the FileLoaderStep tasklet, not in a reader/writer based Step, so then we should change the VcfHeaderReader as well

cyenyxe: The problem is that, if any of those 2 are ever integrated with a chunk-based step (and I see the `PedReader` needing a processor and/or writer to convert to the database schema) they won't behave correctly. I think it's worth changing both to ensure they will work well with tasklets **and** chunk-based steps.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/PedReader.java
```diff
@@ -0,0 +1,62 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.batch.item.ItemReader;
+
+/**
+ * ItemReader that parses a PED file
+ * <p>
+ * PED specs
+ * http://pngu.mgh.harvard.edu/~purcell/plink/data.shtml#ped
+ */
+public class PedReader implements ItemReader<Pedigree> {
```
jorizci: This reader is making a resource acquisition, open and close operations besides the read operation. Please change the reader to an appropriate ResourceAwareItemReaderItemStream or any of it's implementations or extensions that fit better for this use case.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfHeaderReader.java
```diff
@@ -73,13 +75,19 @@ public VcfHeaderReader(File file, VariantSource source) {
      */
     @Override
     public VariantSourceEntity read() throws Exception {
-        VariantVcfReader reader = new VariantVcfReader(source, file.getPath());
-        reader.open();
-        reader.pre();
-        reader.post();
-        reader.close();
+        if (readAlreadyDone) {
```
jorizci: Same comment applies to this reader.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/PedReader.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+        return pedigreeReader.read().get(0);
+    }
+
+    @Override
+    public void open(ExecutionContext executionContext) throws ItemStreamException {
+        checkResourceIsProvided();
+        String resourcePath = getResourcePath();
+        pedigreeReader = new PedigreePedReader(resourcePath);
+        doOpen(resourcePath);
```
jorizci: If the reader is reused, open will be called again but this reader will never read unless you reset the "read finished". I think this applies also to the other reader.

PR #83, opened 2017-02-08
EVA-634 Refactor of MongoConfiguration

jorizci:
MongoConfiguration now is a configuration and is also used with autowiring which is not consistent.

MongoConfiguration now has been refactored so that MongoOperation is injected into the spring context with StepScope. This way the configuration of reader and writers can be simplified. The code has been separated so that we can use static methods from MongoConfigurations and use them in tests where we need to create specific conections.


cyenyxe:
Once the build works again, this looks good to me.

src/main/java/uk/ac/ebi/eva/pipeline/configuration/readers/VariantAnnotationReaderConfiguration.java
```diff
@@ -27,6 +28,7 @@
 /**
  * Configuration to inject a AnnotationFlatFileReader as a Variant Annotation Reader in the pipeline.
  */
+@Component
```
jmmut: not to be nitpicking, but wouldn't it be good to be consistent? I mean, in the future it might not be clear why is this a `@component`, while the other readers and writers are `@Configuration`

jorizci: @Component, what @Component?

PR #84, opened 2017-02-08
updating to biodata 0.4.6 and opencga 0.5.3

jmmut:


install-dependencies.sh
```diff
@@ -1,13 +1,13 @@
 #!/usr/bin/env sh
 
-git clone -b hotfix/0.5 https://github.com/opencb/opencga.git
-cd opencga
-mvn install -DskipTests
-cd ..
-
 git clone https://github.com/EBIvariation/biodata.git
 cd biodata
 git checkout hotfix/0.4
```
cyenyxe: This could be `git clone -b hotfix/0.4 https://github.com/EBIvariation/biodata.git`, same as for the OpenCGA repository.

PR #85, opened 2017-02-15
Refactored JobOptions to remove all innecesary injections and elements

jorizci:
Removed all unused elements. 
Moved configuration variables that were in variant pipeline that were only used as boolean elements to JobOptions getters.
Moved chunk size and restartability from pipelineOptions to JobOptionsGetters.
PipelineOptions only retains the skip tests pending further refactoring.

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/listeners/VariantOptionsConfigurerListener.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/listeners/VariantOptionsConfigurerListener.java
@@ -33,7 +33,6 @@ public class VariantOptionsConfigurerListener implements JobExecutionListener {
 
     private static final Logger logger = LoggerFactory.getLogger(VariantOptionsConfigurerListener.class);
     private final boolean includeSamples;
-    private final boolean compressGenotypes;
     private final boolean calculateStats;
     private final boolean includeStats;
 
@@ -41,11 +40,9 @@ public class VariantOptionsConfigurerListener implements JobExecutionListener {
     private JobOptions jobOptions;
 
     public VariantOptionsConfigurerListener(boolean includeSamples,
-                                            boolean compressGenotypes,
                                             boolean calculateStats,
                                             boolean includeStats) {
         this.includeSamples = includeSamples;
-        this.compressGenotypes = compressGenotypes;
         this.calculateStats = calculateStats;
         this.includeStats = includeStats;
     }
@@ -53,7 +50,7 @@ public class VariantOptionsConfigurerListener implements JobExecutionListener {
     @Override
     public void beforeJob(JobExecution jobExecution) {
         logger.debug("Setting up job " + jobExecution.getJobInstance().getJobName());
-        jobOptions.configureGenotypesStorage(includeSamples, compressGenotypes);
+        jobOptions.configureGenotypesStorage(includeSamples);
         jobOptions.configureStatisticsStorage(calculateStats, includeStats);
     }
 
--- a/src/main/java/uk/ac/ebi/eva/pipeline/parameters/JobOptions.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/parameters/JobOptions.java
@@ -15,27 +15,16 @@
  */
 package uk.ac.ebi.eva.pipeline.parameters;
 
-import org.opencb.biodata.models.variant.VariantSource;
-import org.opencb.biodata.models.variant.VariantStudy;
 import org.opencb.datastore.core.ObjectMap;
 import org.opencb.opencga.lib.common.Config;
 import org.opencb.opencga.storage.core.variant.VariantStorageManager;
-import org.opencb.opencga.storage.core.variant.stats.VariantStatisticsManager;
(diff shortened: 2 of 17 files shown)


src/main/java/uk/ac/ebi/eva/utils/TaskletUtils.java
```diff
@@ -11,13 +11,11 @@
 public class TaskletUtils {
 
     public static TaskletStep generateStep(StepBuilderFactory stepBuilderFactory, String stepName, Tasklet tasklet,
-                                           JobOptions jobOptions) {
+                                           boolean allowStartIfComplete) {
```
jmmut: good idea

src/test/java/uk/ac/ebi/eva/pipeline/ApplicationTest.java
```diff
@@ -43,14 +40,14 @@
     private JobExplorer jobExplorer;
 
     @Autowired
-    private JobOptions jobOptions;
+    private ParametersFromProperties parametersFromProperties;
```
jmmut: why not a DatabaseParameters? I don't like that method `ParametersFromProperties.getDatabaseName()`

jorizci: We are outside of the Job Context, it should faild if we do that.

src/main/java/uk/ac/ebi/eva/utils/TaskletUtils.java
jorizci: The first time we did this abstraction had a lot more of job options that you had to get out... now is just one, which is only a getter... makes more sense to any that want to use the function.

src/test/java/uk/ac/ebi/eva/pipeline/ApplicationTest.java
jorizci: Yeah, tested it fails because of that.

src/main/java/uk/ac/ebi/eva/pipeline/listeners/VariantOptionsConfigurerListener.java
```diff
@@ -33,27 +33,24 @@
...
         this.calculateStats = calculateStats;
         this.includeStats = includeStats;
     }
 
     @Override
     public void beforeJob(JobExecution jobExecution) {
         logger.debug("Setting up job " + jobExecution.getJobInstance().getJobName());
-        jobOptions.configureGenotypesStorage(includeSamples, compressGenotypes);
+        jobOptions.configureGenotypesStorage(includeSamples);
         jobOptions.configureStatisticsStorage(calculateStats, includeStats);
```
cyenyxe: Same as with `includeSamples` and `compressGenotypes`, I think the pair `calculateStats` and `includeStats` always have the same value: if statistics are calculated then they are "included" in the database and vice-versa. But maybe we don't want to get into such a refactoring for this PR.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/JobOptions.java
```diff
@@ -113,186 +65,49 @@ public void loadArgs() throws IOException {
...
-        pipelineOptions.put(JobParametersNames.DB_COLLECTIONS_FILES_NAME, dbCollectionFilesName);
-        pipelineOptions.put(JobParametersNames.DB_COLLECTIONS_FEATURES_NAME, dbCollectionGenesName);
-        pipelineOptions.put(JobParametersNames.DB_COLLECTIONS_STATISTICS_NAME, dbCollectionStatsName);
-        pipelineOptions.put(JobParametersNames.CONFIG_DB_HOSTS, dbHosts);
-        pipelineOptions.put(JobParametersNames.CONFIG_DB_AUTHENTICATIONDB, dbAuthenticationDb);
-        pipelineOptions.put(JobParametersNames.CONFIG_DB_USER, dbUser);
-        pipelineOptions.put(JobParametersNames.CONFIG_DB_PASSWORD, dbPassword);
-        pipelineOptions.put(JobParametersNames.CONFIG_DB_READPREFERENCE, readPreference);
         pipelineOptions.put(JobParametersNames.ANNOTATION_SKIP, skipAnnot);
         pipelineOptions.put(JobParametersNames.STATISTICS_SKIP, skipStats);
```
cyenyxe: Aren't we providing these 2 via job parameters, or is it that the deciders still use `JobOptions`?

src/main/java/uk/ac/ebi/eva/pipeline/listeners/VariantOptionsConfigurerListener.java
jorizci: They have their own issue.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/JobOptions.java
jorizci: Skip Annot and Skip Stats not yet, I'm checking if I can do it and I'll send it in another PR

PR #86, opened 2017-02-15
EVA-644 - JobParamsCornerCases

poggio84:
- JobParametersValidator objects connected into Job objects;
- Extra checks added for non ascii characters and line separator;

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/parameters/ParametersFromProperties.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/parameters/ParametersFromProperties.java
@@ -62,11 +62,38 @@ public class ParametersFromProperties {
     @Value(PROPERTY + JobParametersNames.OUTPUT_DIR_STATISTICS + OR_NULL)
     private String outputDirStats;
 
+    @Value(PROPERTY + JobParametersNames.INPUT_STUDY_NAME + OR_NULL)
+    private String studyName;
+
+    @Value(PROPERTY + JobParametersNames.INPUT_STUDY_TYPE + OR_NULL)
+    private String studyType;
+
+    @Value(PROPERTY + JobParametersNames.OUTPUT_DIR_ANNOTATION + OR_NULL)
+    private String outputDirAnnotation;
+
+    @Value(PROPERTY + JobParametersNames.APP_VEP_CACHE_PATH + OR_NULL)
+    private String vepCachePath;
+
+    @Value(PROPERTY + JobParametersNames.APP_VEP_CACHE_SPECIES + OR_NULL)
+    private String vepCacheSpecies;
+
+    @Value(PROPERTY + JobParametersNames.APP_VEP_CACHE_VERSION + OR_NULL)
+    private String vepCacheVersion;
+
+    @Value(PROPERTY + JobParametersNames.INPUT_FASTA + OR_NULL)
+    private String inputFasta;
+
+    @Value(PROPERTY + JobParametersNames.APP_VEP_NUMFORKS + OR_NULL)
+    private String vepNumForks;
+
+    @Value(PROPERTY + JobParametersNames.APP_VEP_PATH + OR_NULL)
+    private String vepPath;
+
     @Value(PROPERTY + JobParametersNames.STATISTICS_SKIP + OR_NULL)
-    private String statisticsSkip;
+    private String skipStats;
 
     @Value(PROPERTY + JobParametersNames.ANNOTATION_SKIP + OR_NULL)
-    private String annotationSkip;
+    private String skipAnnot;
 
     public Properties getProperties() {
         Properties properties = new Properties();
--- a/src/main/java/uk/ac/ebi/eva/pipeline/parameters/ParametersFromProperties.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/parameters/ParametersFromProperties.java
@@ -90,10 +90,10 @@ public class ParametersFromProperties {
     private String vepPath;
 
     @Value(PROPERTY + JobParametersNames.STATISTICS_SKIP + OR_NULL)
-    private String skipStats;
+    private String statisticsSkip;
 
     @Value(PROPERTY + JobParametersNames.ANNOTATION_SKIP + OR_NULL)
-    private String skipAnnot;
+    private String annotationSkip;
 
     public Properties getProperties() {
         Properties properties = new Properties();
--- a/src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/ParametersValidatorUtil.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/ParametersValidatorUtil.java
@@ -55,7 +55,7 @@ public class ParametersValidatorUtil {
     }
 
     static void checkLength(String stringToValidate, String jobParametersName) throws JobParametersInvalidException {
-        if(stringToValidate.length()>=250){
+        if (stringToValidate.length() >= 250) {
             throw new JobParametersInvalidException(
                     String.format("%s in %s can't exceed 250 characters", stringToValidate, jobParametersName));
         }
--- a/src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/ParametersValidatorUtil.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/ParametersValidatorUtil.java
@@ -38,7 +38,7 @@ public class ParametersValidatorUtil {
 
     static void checkDoesNotContainPrintableCharacters(String stringToValidate,
                                                        String jobParametersName) throws JobParametersInvalidException {
-        Pattern regex = Pattern.compile("[\\p{C}&&[^\n]]");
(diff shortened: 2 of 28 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/ParametersValidatorUtil.java
```diff
@@ -28,6 +28,30 @@
...
+    static void checkIsValidString(String stringToValidate,
+                                   String jobParametersName) throws JobParametersInvalidException {
+        checkIsNotNullOrEmptyString(stringToValidate, jobParametersName);
+        checkAsciiString(stringToValidate, jobParametersName);
+        checkSingleLineString(stringToValidate, jobParametersName);
+    }
+
+    static void checkAsciiString(String stringToValidate,
+                                 String jobParametersName) throws JobParametersInvalidException {
+        if (!stringToValidate.matches("\\A\\p{ASCII}*\\z")) {
```
cyenyxe: We should also support non-ASCII characters, like letters with tildes or dieresis. The ones we should reject are non-printable characters.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/ParametersValidatorUtil.java
```diff
@@ -28,6 +28,30 @@
...
+            throw new JobParametersInvalidException(
+                    String.format("%s in %s contains non ascii characters", stringToValidate, jobParametersName));
+        }
+    }
+
+    static void checkSingleLineString(String stringToValidate,
+                                      String jobParametersName) throws JobParametersInvalidException {
+        final String NEW_LINE = System.getProperty("line.separator");
```
cyenyxe: This should be valid:

```
input.study.title="1000 Genomes Phase 3\nVersion 5"
```

But the one below should fail. Maybe Spring will raise an error automatically because one of the parameters is not a key-value pair.

```
input.study.title=1000 Genomes Phase 3
Version 5
```

But if the same multi-line string is surrounded by quotes it should be fine. Does the pipeline work in a case like this?

```
input.study.title="1000 Genomes Phase 3
Version 5b"
```

src/test/java/uk/ac/ebi/eva/pipeline/parameters/validation/InputStudyNameValidatorTest.java
```diff
@@ -62,8 +62,8 @@ public void inputStudyNameWithDigitsAndWhiteSpacesIsValid() throws JobParameters
         validator.validate(jobParametersBuilder.toJobParameters());
     }
 
-    @Test
-    public void inputStudyNameWithSymbolsIsValid() throws JobParametersInvalidException {
+    @Test(expected = JobParametersInvalidException.class)
+    public void inputStudyNameWithSymbolsIsNotValid() throws JobParametersInvalidException {
```
cyenyxe: These should still be valid.

src/test/java/uk/ac/ebi/eva/pipeline/parameters/validation/ParametersValidatorUtilTest.java
```diff
@@ -54,6 +54,22 @@ public void stringIsNull() throws JobParametersInvalidException {
         ParametersValidatorUtil.checkIsNotNullOrEmptyString(null, JOB_PARAMETER_NAME);
     }
 
+    @Test(expected = JobParametersInvalidException.class)
+    public void stringWithAsciiCharacter() throws JobParametersInvalidException {
```
cyenyxe: Maybe this method should be called `stringWithNonAsciiCharacter` (and the one below should be renamed too), but in any case a string like this should be supported.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/ParametersValidatorUtil.java
jorizci: Btw, as I put in the issue...

Spring batch default databases have some hard limits to job parameters.
* Job parameter id can't have more than 100 characters
* Job parameter string value can't exceed 250 characters

While the first is not a problem, The second should be added as a character limitation in the validation to avoid throwing a big nasty error with database table information.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/ParametersFromProperties.java
```diff
@@ -62,11 +62,38 @@
...
+
+    @Value(PROPERTY + JobParametersNames.APP_VEP_NUMFORKS + OR_NULL)
+    private String vepNumForks;
+
+    @Value(PROPERTY + JobParametersNames.APP_VEP_PATH + OR_NULL)
+    private String vepPath;
+
     @Value(PROPERTY + JobParametersNames.STATISTICS_SKIP + OR_NULL)
-    private String statisticsSkip;
+    private String skipStats;
```
jorizci: This is a regression.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/ParametersFromProperties.java
```diff
@@ -62,11 +62,38 @@
...
 
     @Value(PROPERTY + JobParametersNames.ANNOTATION_SKIP + OR_NULL)
-    private String annotationSkip;
+    private String skipAnnot;
```
jorizci: Same applies

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/ParametersValidatorUtil.java
```diff
@@ -22,12 +22,30 @@
...
+    static void checkIsValidString(String stringToValidate,
+                                   String jobParametersName) throws JobParametersInvalidException {
+        checkIsNotNullOrEmptyString(stringToValidate, jobParametersName);
+        checkNonPrintableCharacters(stringToValidate, jobParametersName);
+        checkLength(stringToValidate, jobParametersName);
+    }
+
+    static void checkNonPrintableCharacters(String stringToValidate,
+                                            String jobParametersName) throws JobParametersInvalidException {
+        Pattern regex = Pattern.compile("[\\p{C}&&[^\n]]");
```
cyenyxe: Can we escape `\n` and `\r` for cross-OS compatibility?

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/ParametersValidatorUtil.java
```diff
@@ -36,6 +54,13 @@ static void checkIsNotNullOrEmptyString(String stringToValidate,
         }
     }
 
+    static void checkLength(String stringToValidate, String jobParametersName) throws JobParametersInvalidException {
+        if(stringToValidate.length()>=250){
```
cyenyxe: White-spaces missing, please reformat.

src/test/java/uk/ac/ebi/eva/pipeline/parameters/validation/InputStudyNameValidatorTest.java
```diff
@@ -63,7 +63,7 @@ public void inputStudyNameWithDigitsAndWhiteSpacesIsValid() throws JobParameters
     }
 
     @Test
-    public void inputStudyNameWithSymbolsIsValid() throws JobParametersInvalidException {
+    public void inputStudyNameWithSymbolsIsNotValid() throws JobParametersInvalidException {
```
cyenyxe: This should be renamed to `inputStudyNameWithSymbolsIsValid` because the exception is no longer expected.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/ParametersValidatorUtil.java
```diff
@@ -22,12 +22,30 @@
...
 public class ParametersValidatorUtil {
 
+    static void checkIsValidString(String stringToValidate,
+                                   String jobParametersName) throws JobParametersInvalidException {
+        checkIsNotNullOrEmptyString(stringToValidate, jobParametersName);
+        checkNonPrintableCharacters(stringToValidate, jobParametersName);
+        checkLength(stringToValidate, jobParametersName);
+    }
+
+    static void checkNonPrintableCharacters(String stringToValidate,
```
cyenyxe: `checkDoesNotContainPrintableCharacters` to match the full sentence structure used in the rest of the class.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/ParametersValidatorUtil.java
```diff
@@ -36,6 +54,13 @@ static void checkIsNotNullOrEmptyString(String stringToValidate,
         }
     }
 
+    static void checkLength(String stringToValidate, String jobParametersName) throws JobParametersInvalidException {
```
cyenyxe: It would be more consistent to check also for length 0 (empty string) in this method, and leave the null only for that specific value.

PR #87, opened 2017-02-15
Refactored skip parameters and removed them from jobOptions.

jorizci:
SkipStepDecider now receives a jobParameterName that is queried to decide if it has to skip the next step. This is done without @StepScopes or @JobScopes as it seems this class was being executed by a thread without the bean scope. Thus we use the JobExecution that the system passes in the decide function to retrieve the value of the parameter.

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/SkipStepDecider.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/SkipStepDecider.java
@@ -15,7 +15,6 @@
  */
 package uk.ac.ebi.eva.pipeline.jobs.deciders;
 
-import org.opencb.datastore.core.ObjectMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.batch.core.JobExecution;
@@ -29,31 +28,21 @@ import org.springframework.batch.core.job.flow.JobExecutionDecider;
 public class SkipStepDecider implements JobExecutionDecider {
     private static final Logger logger = LoggerFactory.getLogger(SkipStepDecider.class);
 
-    private String skipStep;
-    private ObjectMap pipelineOptions;
-
     public static final String SKIP_STEP = "SKIP_STEP";
     public static final String DO_STEP = "DO_STEP";
 
-    /**
-     * @param pipelineOptions ObjectMap that will have a boolean for .get(skipStep), telling whether to skip or not
-     * @param skipStep        name of the key that the user sets to skip a step, e.g. "annotation.create.skip"
-     *                        It's recommended to use pre-defined constants, such as AnnotationJob.SKIP_ANNOT to avoid
-     *                        misspelling mistakes.
-     */
-    public SkipStepDecider(ObjectMap pipelineOptions, String skipStep) {
-        this.skipStep = skipStep;
-        this.pipelineOptions = pipelineOptions;
+    public final String jobParameterName;
+
+    public SkipStepDecider(String jobParameterName){
+        this.jobParameterName = jobParameterName;
     }
 
     @Override
     public FlowExecutionStatus decide(JobExecution jobExecution, StepExecution stepExecution) {
-        if (Boolean.parseBoolean(pipelineOptions.getString(skipStep))) {
-            logger.info("Skipping step because {} is enabled", skipStep);
+        if (Boolean.parseBoolean(jobExecution.getJobParameters().getString(jobParameterName))) {
+            logger.info("Step skipped due to setting {} enabled", jobParameterName);
             return new FlowExecutionStatus(SKIP_STEP);
         }
-
-        logger.info("Running step because {} is disabled", skipStep);
         return new FlowExecutionStatus(DO_STEP);
     }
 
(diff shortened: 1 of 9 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/parameters/ParametersFromProperties.java
```diff
@@ -62,6 +62,12 @@
     @Value(PROPERTY + JobParametersNames.OUTPUT_DIR_STATISTICS + OR_NULL)
     private String outputDirStats;
 
+    @Value(PROPERTY + JobParametersNames.STATISTICS_SKIP + OR_NULL)
+    private String statisticsSkip;
+
+    @Value(PROPERTY + JobParametersNames.ANNOTATION_SKIP + OR_NULL)
```
jmmut: will we need to add these two options to any other class? like InputParameters

jorizci: No, they have they are only use on the skip deciders that can't be injected at all with step or job scope and are readed from the context passed to the decide function. This is only the mandatory integration test property to parameters thingy.

src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/SkipStepDecider.java
```diff
@@ -29,31 +28,21 @@
...
+    public SkipStepDecider(String jobParameterName){
+        this.jobParameterName = jobParameterName;
     }
 
     @Override
     public FlowExecutionStatus decide(JobExecution jobExecution, StepExecution stepExecution) {
-        if (Boolean.parseBoolean(pipelineOptions.getString(skipStep))) {
-            logger.info("Skipping step because {} is enabled", skipStep);
+        if (Boolean.parseBoolean(jobExecution.getJobParameters().getString(jobParameterName))) {
+            logger.info("Step skipped due to setting {} enabled", jobParameterName);
```
cyenyxe: `set [something] enabled` is not correct in English. You could say the following:

```
Step skipped due to {} being enabled
````

File: src/main/java/uk/ac/ebi/eva/pipeline/parameters/ParametersFromProperties.java (hunk 1 above)
Comment (jmmut) [inline src/main/java/uk/ac/ebi/eva/pipeline/parameters/ParametersFromProperties.java]: but did he dieded?

File: src/test/java/uk/ac/ebi/eva/pipeline/jobs/AggregatedVcfJobTest.java [hunk 3]
```diff
@@ -96,6 +96,7 @@ public void aggregatedTransformAndLoadShouldBeExecuted() throws Exception {
                 .inputVcfId("1")
                 .collectionFilesName("files")
                 .timestamp()
+                .annotationSkip(true)
```
cyenyxe: Could this test method check the number of executed steps?

src/main/java/uk/ac/ebi/eva/pipeline/parameters/ParametersFromProperties.java
```diff
@@ -62,6 +62,12 @@
     @Value(PROPERTY + JobParametersNames.OUTPUT_DIR_STATISTICS + OR_NULL)
     private String outputDirStats;
 
+    @Value(PROPERTY + JobParametersNames.STATISTICS_SKIP + OR_NULL)
```
cyenyxe: Should default to `false` instead of `null`.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/ParametersFromProperties.java
cyenyxe: Should default to `false` instead of `null`.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/GenotypedVcfJobWorkflowTest.java
```diff
@@ -236,11 +232,11 @@ private JobParameters initVariantConfigurationJob() throws IOException {
                 .vepCacheSpecies("")
                 .vepCacheVersion("")
                 .vepNumForks("")
-                .vepPath(getResource(MOCK_VEP).getPath());
+                .vepPath(getResource(MOCK_VEP).getPath())
+                .annotationSkip(false)
+                .statisticsSkip(false);
```
cyenyxe: These 2 should default to `false`. Do we want to keep them explicit to ensure this behavior even if the main code changes?

src/main/java/uk/ac/ebi/eva/pipeline/parameters/ParametersFromProperties.java
jorizci: This is only to pass the value from the properties as jobParameters. It is only for Integration test and if no defined in the property file the behaviour defaults to false value.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/GenotypedVcfJobWorkflowTest.java
jorizci: If you don't pass them they default to false

src/main/java/uk/ac/ebi/eva/pipeline/parameters/ParametersFromProperties.java
jorizci: Same comment applies.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/AggregatedVcfJobTest.java
jorizci: The test already checks the executed steps.

PR #88, opened 2017-02-16
EVA-524 Dependencies between tests

jmmut:
Make sure that all the tests are using its own temporary databases and folders.

Also, restructure the test resources folder.


src/test/java/uk/ac/ebi/eva/pipeline/io/readers/AggregatedVcfReaderTest.java
```diff
@@ -31,11 +32,14 @@
 
     private static final String STUDY_ID = "7";
 
-    private static final String INPUT_FILE_PATH = "/aggregated.vcf.gz";
+    private static final String INPUT_FILE_PATH = "/input-files/aggregated.vcf.gz";
```
cyenyxe: If there are new folders like `/input-files/ped` and `/input-files/statistics`, then it would be more consistent to keep the VCF files in `/input-files/vcf`.

PR #92, opened 2017-02-20
EVA-556 StepLauncherTestUtils

jorizci:
-Added StepLauncherTestUtils.java to test steps without requiring the injection of a full job.
-Modified step tests to use the new test launcher.

### DIFF
--- a/src/test/java/uk/ac/ebi/eva/test/configuration/BatchTestConfiguration.java
+++ b/src/test/java/uk/ac/ebi/eva/test/configuration/BatchTestConfiguration.java
@@ -1,12 +1,17 @@
 package uk.ac.ebi.eva.test.configuration;
 
+import org.springframework.batch.core.Job;
+import org.springframework.batch.core.Step;
 import org.springframework.batch.core.configuration.annotation.EnableBatchProcessing;
 import org.springframework.batch.test.JobLauncherTestUtils;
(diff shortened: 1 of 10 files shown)


src/test/java/uk/ac/ebi/eva/test/configuration/BatchTestConfiguration.java
```diff
@@ -15,8 +20,16 @@
...
 
     @Bean
+    @ConditionalOnBean(value = {Job.class})
     public JobLauncherTestUtils jobLauncherTestUtils() {
         return new JobLauncherTestUtils();
     }
 
+    @Bean
+    @ConditionalOnBean(value = {Step.class})
+    @ConditionalOnMissingBean(value = {Job.class})
```
jmmut: Cool annotation, but were these two `@ConditionalOnBean` needed?

jorizci: Yes, they are totally required. This makes that if no job is found in the configuration of the test then we don't inject the JobLauncherTestUtils which would fail if no Job is found (Or it will throw a dependency injection error) , and then the StepLauncher would not be injected if no Step is found (Or it will throw a dependency injection error) or if a Job is found the context in which case you want to use the JobLauncherTestUtils not the StepLauncherTestUtils

tcezard:
Not Needed: closing

PR #95, opened 2017-02-24
EVA-673 Drop variants reported only by a given study

jmmut:


src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/tasklets/SingleStudyVariantsDropperStep.java
```diff
@@ -0,0 +1,73 @@
...
+    public RepeatStatus execute(StepContribution contribution, ChunkContext chunkContext) throws Exception {
+
+        String filesStudyIdField = String.format("%s.%s", FILES_FIELD, STUDYID_FIELD);
+        Query query = new Query(
+                new Criteria(filesStudyIdField).is(inputParameters.getStudyId())
+                        .and(FILES_FIELD).size(1));
+
+        logger.info("removing single study variants. used query: {}", query);
+        WriteResult writeResult = mongoOperations.remove(query, dbParameters.getCollectionVariantsName());
+        logger.info("result: {}", writeResult.toString());
```
jorizci: I think these two should be debug more than info.

src/main/java/uk/ac/ebi/eva/pipeline/jobs/DropStudyJob.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Bean(DROP_STUDY_JOB)
+    @Scope("prototype")
+    public Job dropStudyJob(JobBuilderFactory jobBuilderFactory) {
+        logger.debug("Building '" + DROP_STUDY_JOB + "'");
+
+        JobBuilder jobBuilder = jobBuilderFactory
+                .get(DROP_STUDY_JOB)
+                .incrementer(new RunIdIncrementer())
+                .validator(new AggregatedVcfJobParametersValidator());
```
jorizci: [Waves hand]This is not the validator you're looking for...

src/main/java/uk/ac/ebi/eva/pipeline/jobs/DropStudyJob.java
```diff
@@ -0,0 +1,70 @@
...
+
+import uk.ac.ebi.eva.pipeline.jobs.steps.DropSingleStudyVariantsStep;
+
+import static uk.ac.ebi.eva.pipeline.configuration.BeanNames.DROP_SINGLE_STUDY_VARIANTS_STEP;
+import static uk.ac.ebi.eva.pipeline.configuration.BeanNames.DROP_STUDY_JOB;
+
+/**
+ * Job that removes a study from the database. Given a study to remove:
+ * <p>
+ * remove single study variants --> pull study entries in the rest of variants --> remove file entry in files collection
```
cyenyxe: "remove variants in single study"

"pull study entries from"

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/tasklets/SingleStudyVariantsDropperStep.java
```diff
@@ -0,0 +1,72 @@
...
+ * Tasklet that removes from mongo all the variants that have only one entry from a given study to delete.
+ * If we removed the entry instead of the whole variant, we could end up keeping variants that don't appear in any
+ * study (i.e. an empty "files" array in the variant mongo document), which doesn't make sense, so we remove the
+ * complete document for those cases first.
+ * <p>
+ * Input: a studyId
+ * <p>
+ * Output: those variants are removed
+ */
+public class SingleStudyVariantsDropperStep implements Tasklet {
```
cyenyxe: I am extremely keen on renaming the classes in this package to `XYZTasklet` instead of `XYZStep`. That will reduce the confusion with the configuration classes that inject `TaskletStep` objects, and we won't need to use names with kind of an opposite structure. Instead of:

* DropSingleStudyVariantsStep
* SingleStudyVariantsDropperStep

We could have:

* DropSingleStudyVariantsStep
* DropSingleStudyVariantsTasklet

src/test/java/uk/ac/ebi/eva/pipeline/jobs/steps/DropSingleStudyVariantsStepTest.java
```diff
@@ -0,0 +1,134 @@
...
+    public void testNoVariantsToDrop() throws Exception {
+        String databaseName = insertDocuments(COLLECTION_VARIANTS_NAME, Arrays.asList(
+                VariantData.getVariantWithOneStudy(),
+                VariantData.getVariantWithTwoStudies()));
+
+        checkDrop(databaseName, EXPECTED_VARIANTS_AFTER_DROP_STUDY);
+    }
+
+    @Test
+    public void testDropMatchingSingleStudyVariant() throws Exception {
```
cyenyxe: `testOneVariantToDrop`, to match the naming of other test methods?

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/tasklets/SingleStudyVariantsDropperStep.java
jmmut: I agree on the bad naming, but suggest that the problem is not that one is a step and the other a tasklet, but that one is a step (or a tasklet) and the other is the configuration class that injects it. I'd say this is more informative:
- DropSingleStudyVariantsStep
- DropSingleStudyVariantsStepConfiguration

and if we could move those _Configuration under the configuration folder, it would be great

jorizci: Yeah, they are configuration classes and they should be moved to their propper place, still, the XYZTasklet is also reasonable.

I propose to perform the suggested naming change of this class but maintain it as the current "standard" and submit a secont pr with the change names / package rearangements to avoid stress inducing rebases.

PR #96, opened 2017-02-24
Architecture improvements - More Spring Batch compliant

cyenyxe:
To make test run completely independent from each other, it was necessary to:

* Ensure they used different database names and test files
* Parameters from one job weren't reused across multiple ones

In order to do so, we conducted the following work that, among other things, involved migrating to a more Spring-Batch compliant architecture:

* Use a random Mongo database for each test case
* Use a random test folder (and output files contained within it) for each test case
* Replace our custom job parameters handler to another based on the Spring Batch built-in `JobParameters` class
* Fully validate parameters (numbers, strings, file paths, etc.) using a myriad of `JobParametersValidator` :wink:
* Implement a custom `JobLauncherCommandLineRunner` that allows to read parameters from a properties file in addition to the CLI


src/test/java/uk/ac/ebi/eva/pipeline/jobs/GenotypedVcfJobWorkflowTest.java
```diff
@@ -16,82 +16,72 @@
...
+    private JobLauncherTestUtils jobLauncherTestUtils;
 
-    private String vepOutput;
+    private static String opencgaHome = System.getenv("OPENCGA_HOME") != null ? System
+            .getenv("OPENCGA_HOME") : "/opt/opencga";
 
-    private static String opencgaHome = System.getenv("OPENCGA_HOME") != null ? System.getenv("OPENCGA_HOME") :
-            "/opt/opencga";
+    @Autowired
+    private JobOptions jobOptions;  // we need this for stats.skip and annot.skip
```
jmmut: things like this are not true anymore. in the near future there will be yet another tests cleanup

jorizci: Yeah, this can be removed. Could we try to pass a code checking tool?

PR #98, opened 2017-02-27
fix bug: using aggregate mapping from resources

jmmut:
it seems it's not possible to use a resource from within the jar as a regular file: http://stackoverflow.com/a/18055478/2375586

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/utils/FileUtils.java
+++ b/src/main/java/uk/ac/ebi/eva/utils/FileUtils.java
@@ -22,7 +22,6 @@ import org.springframework.core.io.Resource;
 import uk.ac.ebi.eva.pipeline.io.GzipLazyResource;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/ac/ebi/eva/utils/FileUtils.java
```diff
@@ -66,10 +65,13 @@ public static File getResource(String resourcePath) {
         return new File(FileUtils.class.getResource(resourcePath).getFile());
     }
 
-    public static Properties getPropertiesFile(String propertiesFilePath) throws IOException {
-        InputStream input = new FileInputStream(propertiesFilePath);
+    public static InputStream getResourceAsStream(String resourcePath) {
+        return FileUtils.class.getResourceAsStream(resourcePath);
```
cyenyxe: `class` shouldn't be necessary, see usage without it in `VariantVcfExacFactory.java`.

jmmut: if I remove the class, it becomes a recursive method XD

jorizci: Even if you change the name you need the class to call that method.

PR #100, opened 2017-03-02
EVA-658 Read non-annotated variants only for study being loaded

jmmut:
the main jobs (aggregated and genotyped) require a studyId to be specified, so the final flow of annotation will only annotate that study.

The annotation job doesn't require a study (the validators are not made yet), so if a study is provided, only that study will be annotated. if the study is an empty string, all studies will be annotated.


src/main/java/uk/ac/ebi/eva/utils/MongoDBHelper.java
```diff
@@ -86,8 +86,4 @@ public static String buildStorageId(String chromosome, int start, String referen
         return builder.toString();
     }
 
-    public static String buildStorageFileId(String studyId, String fileId) {
```
cyenyxe: I would suggest leaving the code split in `MongoDBHelper` and `URLHelper`, as they previously were. The fact that their implementations currently match doesn't mean they are the same, but they apply to very different contexts. It is not intuitive to look in `URLHelper` if to get a Mongo key.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
```diff
@@ -31,13 +33,22 @@
...
+     * @param studyId nullable. If null, the query brings all non-annotated variants in the collection.
+     *  If not null, bring only non-annotated variants from that study.
+     */
+    public NonAnnotatedVariantsMongoReader(MongoOperations template, String collectionsVariantsName, String studyId) {
         setTemplate(template);
         setCollection(collectionsVariantsName);
 
-        DBObject query = BasicDBObjectBuilder.start().add("annot.ct.so", new BasicDBObject("$exists", false)).get();
+        BasicDBObjectBuilder queryBuilder = BasicDBObjectBuilder.start();
+        if (!studyId.isEmpty()) {
```
cyenyxe: This will break if `studyId` is null. Please also add the corresponding test case.

src/main/java/uk/ac/ebi/eva/utils/MongoDBHelper.java
jmmut: I moved that method because has nothing to do with mongo. The underscore is only used as the VariantSourceEntry key in the Variant object and in some files like the stats one: "sid_fid.variants.stats.json"

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
```diff
@@ -37,10 +38,17 @@
...
-     *  If not null, bring only non-annotated variants from that study.
+     * @param studyId Can be the empty string, meaning to bring all non-annotated variants in the collection.
+     *  If the studyId string is not empty, bring only non-annotated variants from that study. This parameter should
+     *  not be null in any case.
      */
-    public NonAnnotatedVariantsMongoReader(MongoOperations template, String collectionsVariantsName, String studyId) {
+    public NonAnnotatedVariantsMongoReader(MongoOperations template, String collectionsVariantsName, String studyId)
+            throws JobParametersInvalidException {
+        if (studyId == null) {
+            throw new JobParametersInvalidException("NonAnnotatedVariantsMongoReader needs a non-null studyId " +
```
cyenyxe: The exception type usually thrown with incorrect arguments is `IllegalArgumentException`.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/steps/VepInputGeneratorStepTest.java
```diff
@@ -73,15 +80,24 @@
 
     @Test
     public void shouldGenerateVepInput() throws Exception {
+        checkVepInputWasGenerated(STUDY_ID);
```
poggio84: I prefer to have the assert in the test method.. but this is just my personal preference. I found it easier to understand even if I have to add a bit of code duplication.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/steps/VepInputGeneratorStepTest.java
```diff
@@ -73,15 +80,24 @@
 
     @Test
     public void shouldGenerateVepInput() throws Exception {
```
poggio84: shouldGenerateVepInputForStudyId?

src/test/java/uk/ac/ebi/eva/pipeline/jobs/steps/VepInputGeneratorStepTest.java
```diff
@@ -93,5 +109,6 @@ public void shouldGenerateVepInput() throws Exception {
 
         assertTrue(vepInput.exists());
         assertEquals("20\t60343\t60343\tG/A\t+", readFirstLine(vepInput));
+        assertEquals(EXPECTED_NON_ANNOTATED_VARIANTS, getLines(new FileInputStream(vepInput)));
     }
```
poggio84: Here the same method checkVepInputWasGenerated is called twice with different inputStudyId but then you are always checking the same result?

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
```diff
@@ -31,13 +34,29 @@
...
+            + VariantSourceEntryToDBObjectConverter.STUDYID_FIELD;
+
+    /**
+     * @param studyId Can be the empty string, meaning to bring all non-annotated variants in the collection.
+     *  If the studyId string is not empty, bring only non-annotated variants from that study. This parameter should
+     *  not be null in any case.
+     */
+    public NonAnnotatedVariantsMongoReader(MongoOperations template, String collectionsVariantsName, String studyId)
+            throws JobParametersInvalidException {
+        if (studyId == null) {
```
jmmut: this reader may be used without injection into a step, the usage with a validator is not required. I don't see much difference between an assertionError and a RuntimeException.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/steps/VepInputGeneratorStepTest.java
jmmut: yes, the dump only contains one study, so the result is the same. The second test is more to check that the parameters validator allow an empty string as studyId. I will change the test name to highlight that.

jmmut: I mean, there's a tricky issue, where we require a non-empty studyId in the genotyped-vcf-job, so only one study will be annotated, but for the annotation-job, we allow both a studyId or an empty string for all studies, and a different stepParametersValidator will need to be done

PR #103, opened 2017-03-06
EVA-680 Job to drop study from Mongo

jmmut:
- job wiring up the three steps (dropSingleStudyVariantsStep, dropVariantsAndStatisticsByStudyStep, dropVariantsAndStatisticsByStudyStep)
- job test
- parameters validator for the job (and validator test)
- parameters validators for the three steps (and validator tests)
- some general improvements (extracting common methods and improving logs)

### DIFF
--- a/src/test/java/uk/ac/ebi/eva/test/utils/DropStudyJobTestUtils.java
+++ b/src/test/java/uk/ac/ebi/eva/test/utils/DropStudyJobTestUtils.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2015-2017 EMBL - European Bioinformatics Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.ac.ebi.eva.test.utils;
+
+import com.mongodb.BasicDBObject;
+import com.mongodb.DBCollection;
+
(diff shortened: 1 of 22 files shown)


src/test/java/uk/ac/ebi/eva/test/utils/DropStudyJobTestUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+import com.mongodb.BasicDBObject;
+import com.mongodb.DBCollection;
+
+import static org.junit.Assert.assertEquals;
+import static uk.ac.ebi.eva.commons.models.converters.data.VariantToDBObjectConverter.FILES_FIELD;
+import static uk.ac.ebi.eva.commons.models.converters.data.VariantToDBObjectConverter.STATS_FIELD;
+import static uk.ac.ebi.eva.commons.models.data.VariantSourceEntity.STUDYID_FIELD;
+
+public class DropStudyJobTestUtils {
```
cyenyxe: Given the low complexity of these checks, leaving them in the 2 classes where they are used would make easier to understand what the code is actually testing.

jmmut: I would say that the methods improve code clarity and reduce redundancy, but the main reason to extract them was in fact to highlight the common checks between the job and the steps.

jorizci: I think jmmut has a point here and the code seems clear to me, it seems that is there are future issues there i can grasp the meaning quickly even if the assert functions. The only improvement would change the name to assert instead of check to reassure its purpose.

cyenyxe: Fine then. Could they be accessed in a consistent way? `PullFilesAndStatisticsByStudyStepTest` uses a regular import but `DropStudyJobTest` uses static imports.

cyenyxe:
The code looks good to me know. Can we have a new properties file added to the "examples" folder before merging?

jmmut:
added an extra fix https://github.com/EBIvariation/eva-pipeline/pull/103/commits/4586b5969526c80ef98bd6d7bd22717d048bf9f3

PR #105, opened 2017-03-09
EVA-461 Remove Chr and chr prefix in CHROM column during loading

jmmut:
Note:
- Variant.setChromosome does not remove the "chr" anymore. The only place that should do that is the factory.

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VariantVcfFactory.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VariantVcfFactory.java
@@ -64,8 +64,8 @@ public class VariantVcfFactory {
             throw new IllegalArgumentException("Not enough fields provided (min 8)");
         }
 
-        String chromosome = removeChrPrefix(fields[0]);
-        int position = Integer.parseInt(fields[1]);
+        String chromosome = getChromosomeWithoutPrefix(fields);
+        int position = getPosition(fields);
         Set<String> ids = getIds(fields);
         String reference = getReference(fields);
         String[] alternateAlleles = getAlternateAlleles(fields, chromosome, position, reference);
@@ -108,7 +108,8 @@ public class VariantVcfFactory {
      * Replace "chr" references only at the beginning of the chromosome name.
      * For instance, tomato has SL2.40ch00 and that should be kept that way
      */
-    private String removeChrPrefix(String chromosome) {
+    private String getChromosomeWithoutPrefix(String[] fields) {
+        String chromosome = fields[0];
         boolean ignoreCase = true;
         int startOffset = 0;
         String prefixToRemove = "chr";
@@ -116,6 +117,18 @@ public class VariantVcfFactory {
             return chromosome.substring(prefixToRemove.length());
         }
         return chromosome;
+    }
+
+    private int getPosition(String[] fields) {
+        return Integer.parseInt(fields[1]);
+    }
+
+    private Set<String> getIds(String[] fields) {
+        Set<String> ids = new HashSet<>();
+        if (!fields[2].equals(".")) {    // note!: we store a "." as an empty set, not a set with an empty string
+            ids.addAll(Arrays.asList(fields[2].split(";")));
+        }
+        return ids;
     }
 
     private String getReference(String[] fields) {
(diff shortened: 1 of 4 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VariantVcfFactory.java
```diff
@@ -140,99 +120,68 @@
...
-                String inAlt = alt.substring(indexOfDifference);
-                return new VariantKeyFields(start, end, ref, inAlt);
-            }
-        }
+        indexOfDifference = StringUtils.indexOfDifference(reference, alternate);
+        int start = position + indexOfDifference;
+        int length = max(reference.length(), alternate.length());
+        int end = position + length - 1;    // why -1? because end is inclusive
+        String normalizedReference = reference.substring(indexOfDifference);
+        String normalizedAlternate = alternate.substring(indexOfDifference);
```
cyenyxe: These will always create 2 new strings. This branching could prevent it:

```java
// Assign indexOfDifference, start, length and end the same way for all cases
// Reference and alternate stay the same when indexOfDifference is zero

if (indexOfDifference > 0) {
    reference = reference.substring(indexOfDifference);
    alternate = alternate.substring(indexOfDifference);
}
return new VariantKeyFields(start, end, reference, alternate);
```

src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VariantVcfFactory.java
```diff
@@ -140,99 +120,68 @@
...
      * @param position Input starting position
      * @param reference Input reference allele
-     * @param alt Input alternate allele
+     * @param alternate Input alternate allele
      * @return The new start, end, reference and alternate alleles
      */
-    protected VariantKeyFields createVariantsFromIndelNoEmptyRefAlt(int position, String reference, String alt) {
-        int indexOfDifference;
+    protected VariantKeyFields normalizeSuffixFirst(int position, String reference, String alternate)
+            throws NotAVariantException {
```
jmmut: I think we have a skip listener, and we allow up to 50 variants to fail. without the listener, the first no-variant would make the step fail.

src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VariantVcfFactory.java
```diff
@@ -140,99 +120,68 @@
...
+        if (reference.equals(alternate)) {
+            throw new NotAVariantException();
```
jmmut: I did it this way because the normalization didn't need those parameters. following recommendations for using exceptions, the normalization can not recover by itself, nor decide what to do with that error, so it raises until someone knows the context of the error. admittedly, this is case way simpler, so I will just add the parameters

src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VariantVcfFactory.java
```diff
@@ -140,99 +120,68 @@
...
      *
      * @param position Input starting position
      * @param reference Input reference allele
-     * @param alt Input alternate allele
+     * @param alternate Input alternate allele
      * @return The new start, end, reference and alternate alleles
      */
-    protected VariantKeyFields createVariantsFromIndelNoEmptyRefAlt(int position, String reference, String alt) {
-        int indexOfDifference;
+    protected VariantKeyFields normalizeSuffixFirst(int position, String reference, String alternate)
```
jmmut: agreed normalizeLeftAlign

src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VariantVcfFactory.java
```diff
@@ -62,65 +64,27 @@
             throw new IllegalArgumentException("Not enough fields provided (min 8)");
         }
 
-        List<Variant> variants = new LinkedList<>();
-
-        String chromosome = fields[0];
+        String chromosome = removeChrPrefix(fields[0]);
         int position = Integer.parseInt(fields[1]);
```
cyenyxe: Shouldn't these be `getChromosome` (or `getChromosomeWithoutChrPrefix`) and `getPosition`, for the sake of consistency?

PR #107, opened 2017-03-20
EVA-348 Step to store variant annotation version

jmmut:
This PR adds a step AnnotationMetadataStep to the annotation flow. The step writes a small document represented by AnnotationMetadata into its collection, intended to track VEP versions and VEP cache versions used in that database.

For the VEP version a parameter was added: `app.vep.version`

Also, added the parameter validators for that step (included in the genotyped and aggregated jobs).


src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/VepVersionValidator.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Checks that the VEP version option has benn filled in
+ *
+ * @throws JobParametersInvalidException If the vep version is null or empty
+ */
+public class VepVersionValidator implements JobParametersValidator {
+    @Override
+    public void validate(JobParameters parameters) throws JobParametersInvalidException {
+        ParametersValidatorUtil
+                .checkIsValidString(parameters.getString(JobParametersNames.APP_VEP_VERSION),
```
jmmut: What are we getting if we use integers? I would prefer to stick with strings, as it's understandable that they used something like 88.1 or to note the migration from ensembl-tools to ensembl-vep

src/test/java/uk/ac/ebi/eva/pipeline/jobs/steps/AnnotationMetadataStepTest.java
```diff
@@ -0,0 +1,138 @@
...
+                .toJobParameters();
+
+        JobExecution jobExecution = jobLauncherTestUtils.launchStep(BeanNames.LOAD_ANNOTATION_METADATA_STEP, jobParameters);
+
+        assertEquals(ExitStatus.COMPLETED, jobExecution.getExitStatus());
+        assertEquals(BatchStatus.COMPLETED, jobExecution.getStatus());
+    }
+
+    @Test
+    public void shouldNotAddRedundantVersions() throws Exception {
```
cyenyxe: I suspect the method names are swapped between `shouldNotAddRedundantVersions` and `shouldKeepOtherVersions`, as the former currently inserts 2 different versions and the first tries to insert the same twice. It should be the opposite.

PR #108, opened 2017-03-22
EVA-706 Create Spring Batch writer that takes Variants and annotates into a file

jmmut:
This PR adds the VepAnnotationFileWriter, and belongs to a bigger refactor whose purpose is to have the annotation flow like this:
- annotation flow:
  - GenerateAnnotationStep
    - NonAnnotatedVariantsReader<VariantWrapper>
    - VepAnnotationFileWriter<VariantWrapper>  ---**current PR**---
  - AnnotationLoaderStep
    - VariantAnnotationReader<VariantAnnotation>
    - VariantAnnotationWriter<VariantAnnotation>

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright 2015-2017 EMBL - European Bioinformatics Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.ac.ebi.eva.pipeline.io.writers;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.batch.item.ExecutionContext;
+import org.springframework.batch.item.ItemStreamException;
+import org.springframework.batch.item.ItemStreamWriter;
+
+import uk.ac.ebi.eva.pipeline.model.VariantWrapper;
+import uk.ac.ebi.eva.pipeline.parameters.AnnotationParameters;
+
+import java.io.BufferedInputStream;
+import java.io.BufferedOutputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+public class VepAnnotationFileWriter implements ItemStreamWriter<VariantWrapper> {
+    private static final Logger logger = LoggerFactory.getLogger(VepAnnotationFileWriter.class);
+
+    private AnnotationParameters annotationParameters;
+
+    private Integer chunkSize;
+
+    private Process process;
+
+    private OutputStream perlStdin;
+
+    private final Long timeoutInSeconds;
+
+    public VepAnnotationFileWriter(Long timeoutInSeconds, Integer chunkSize, AnnotationParameters annotationParameters) {
+        this.timeoutInSeconds = timeoutInSeconds;
+        this.chunkSize = chunkSize;
+        this.annotationParameters = annotationParameters;
+    }
+
+    @Override
+    public void open(ExecutionContext executionContext) throws ItemStreamException {
+        ProcessBuilder processBuilder = new ProcessBuilder("perl", annotationParameters.getVepPath(),
+                "--cache",
+                "--cache_version", annotationParameters.getVepCacheVersion(),
+                "-dir", annotationParameters.getVepCachePath(),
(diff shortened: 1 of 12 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        perlStdin = new BufferedOutputStream(process.getOutputStream());
+        InputStream perlStdout = process.getInputStream();
+        String vepOutput = annotationParameters.getVepOutput();
+
+        writingOk = new AtomicBoolean(false);
+        outputCapturer = new Thread(() -> {
+            long written = 0;
+            try (GZIPOutputStream outputStream = new GZIPOutputStream(new FileOutputStream(vepOutput))) {
```
jorizci: Change this call to the Constructor with the append parameter so that we don't overwrite the file over and over and over (as we found)

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
```diff
@@ -0,0 +1,225 @@
...
+
+    private final Long timeoutInSeconds;
+
+    private Thread outputCapturer;
+
+    private AtomicBoolean writingOk;
+
+    private boolean alreadyClosed;
+
+    public VepAnnotationFileWriter(Long timeoutInSeconds, Integer chunkSize, AnnotationParameters annotationParameters) {
```
jorizci: I agrre with Diego, It will be clearer if the concrete logic of VEP would be encapsulated in it's own class, or that when constructed returns a class that we use as interface or any similar solution.

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
```diff
@@ -0,0 +1,225 @@
...
+
+        logger.info("Starting VEP process");
+        process = null;
+        try {
+            process = processBuilder.start();
+        } catch (IOException e) {
+            throw new ItemStreamException(e);
+        }
+
+        perlStdin = new BufferedOutputStream(process.getOutputStream());
```
jorizci: Stdin? We don't charge for letters!

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
```diff
@@ -0,0 +1,225 @@
...
+                        e);
+            }
+            throw new ItemStreamException("Error while running VEP (exit status " + exitValue + "). See "
+                    + errorLog + " for the errors description from VEP.");
+        }
+    }
+
+    private void checkOutputWritingStatus() {
+        try {
+            final long CONVERT_SECONDS_TO_MILLISECONDS = 1000L;
```
jorizci: constant in the class?

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
```diff
@@ -0,0 +1,225 @@
...
+        }
+        if (!writingOk.get()) {
+            throw new ItemStreamException("VEP output writer thread could not finish properly. ");
+        }
+    }
+
+    /**
+     * read all the inputStream and write it into the outputStream
+     */
+    private long connectStreams(InputStream inputStream, OutputStream outputStream) throws IOException {
```
jorizci: I would give a second thought to this method. It's not only that the return value is not used, also, the written only conts the number of successful reads. This will not be equal to the full readed size in some scenarios. Not only that, this method is doing a hungry active waiting. It should sleep from time to time or you are just gonna have a lot of wasted cycles.

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
jmmut: I thought it was easier to follow the logic with `perlStdin` than with `processStandardInput`, (which by the way is actually a outputStream).

Ok it seems only I found it clearer this way :D

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
```diff
@@ -0,0 +1,225 @@
...
+        String vepOutput = annotationParameters.getVepOutput();
+
+        writingOk = new AtomicBoolean(false);
+        outputCapturer = new Thread(() -> {
+            long written = 0;
+            try (GZIPOutputStream outputStream = new GZIPOutputStream(new FileOutputStream(vepOutput))) {
+                written = connectStreams(new BufferedInputStream(perlStdout), outputStream);
+                writingOk.set(true);
+            } catch (IOException  e) {
+                logger.error("Writing the VEP output to " + vepOutput + " failed. ", e);
```
jmmut: we are tracking if the writing was ok with the `AtomicBoolean writingOk` which is always false until we finish writing, just above the catch

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
jmmut: it is used only in the vep output (we log that), but not in the vep error output.

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
jmmut: in the next commit we have split the variantWrapper writing into a string, and the vep process handling, as if it was a regular output stream

PR #109, opened 2017-03-22
EVA-705 Modify NonAnnotatedVariantsReader to return java objects instead of DBObject

jmmut:
This PR improves the NonAnnotatedVariantsReader, and belongs to a bigger refactor whose purpose is to have the annotation flow like this:

- annotation flow:
  - GenerateAnnotationStep
     - NonAnnotatedVariantsReader ---**current PR**---
     - VepAnnotationFileWriter
  - AnnotationLoaderStep
    - VariantAnnotationReader
    - VariantAnnotationWriter

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/configuration/readers/NonAnnotatedVariantsMongoReaderConfiguration.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/configuration/readers/NonAnnotatedVariantsMongoReaderConfiguration.java
@@ -21,6 +21,7 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.data.mongodb.core.MongoOperations;
 
 import uk.ac.ebi.eva.pipeline.io.readers.NonAnnotatedVariantsMongoReader;
+import uk.ac.ebi.eva.pipeline.parameters.AnnotationParameters;
 import uk.ac.ebi.eva.pipeline.parameters.DatabaseParameters;
 import uk.ac.ebi.eva.pipeline.parameters.InputParameters;
 
--- a/src/main/java/uk/ac/ebi/eva/pipeline/configuration/readers/NonAnnotatedVariantsMongoReaderConfiguration.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/configuration/readers/NonAnnotatedVariantsMongoReaderConfiguration.java
@@ -21,7 +21,6 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.data.mongodb.core.MongoOperations;
 
 import uk.ac.ebi.eva.pipeline.io.readers.NonAnnotatedVariantsMongoReader;
-import uk.ac.ebi.eva.pipeline.parameters.AnnotationParameters;
 import uk.ac.ebi.eva.pipeline.parameters.DatabaseParameters;
 import uk.ac.ebi.eva.pipeline.parameters.InputParameters;
 
--- a/src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
@@ -23,9 +23,6 @@ import org.opencb.opencga.storage.mongodb.variant.DBObjectToVariantConverter;
 import org.springframework.batch.item.ExecutionContext;
 import org.springframework.batch.item.ItemStreamException;
 import org.springframework.batch.item.ItemStreamReader;
-import org.springframework.batch.item.NonTransientResourceException;
-import org.springframework.batch.item.ParseException;
-import org.springframework.batch.item.UnexpectedInputException;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.data.mongodb.core.MongoOperations;
 
@@ -90,8 +87,7 @@ public class NonAnnotatedVariantsMongoReader implements ItemStreamReader<Variant
     }
 
     @Override
-    public VariantWrapper read()
-            throws Exception, UnexpectedInputException, ParseException, NonTransientResourceException {
+    public VariantWrapper read() throws Exception {
         DBObject dbObject = delegateReader.read();
         if (dbObject != null) {
             Variant variant = converter.convertToDataModelType(dbObject);
--- a/src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
@@ -20,10 +20,14 @@ import com.mongodb.BasicDBObjectBuilder;
 import com.mongodb.DBObject;
 import org.opencb.biodata.models.variant.Variant;
 import org.opencb.opencga.storage.mongodb.variant.DBObjectToVariantConverter;
-import org.springframework.batch.item.support.AbstractItemCountingItemStreamItemReader;
+import org.springframework.batch.item.ExecutionContext;
+import org.springframework.batch.item.ItemStreamException;
+import org.springframework.batch.item.ItemStreamReader;
+import org.springframework.batch.item.NonTransientResourceException;
+import org.springframework.batch.item.ParseException;
+import org.springframework.batch.item.UnexpectedInputException;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.data.mongodb.core.MongoOperations;
-import org.springframework.util.ClassUtils;
 
 import uk.ac.ebi.eva.commons.models.converters.data.VariantSourceEntryToDBObjectConverter;
 import uk.ac.ebi.eva.commons.models.converters.data.VariantToDBObjectConverter;
@@ -37,8 +41,7 @@ import javax.annotation.PostConstruct;
  * {@link org.springframework.batch.item.data.MongoItemReader} is using
  * pagination and it is slow with large collections
  */
-public class NonAnnotatedVariantsMongoReader extends AbstractItemCountingItemStreamItemReader<VariantWrapper>
-        implements InitializingBean {
+public class NonAnnotatedVariantsMongoReader implements ItemStreamReader<VariantWrapper>, InitializingBean {
 
     private MongoDbCursorItemReader delegateReader;
 
@@ -53,7 +56,6 @@ public class NonAnnotatedVariantsMongoReader extends AbstractItemCountingItemStr
      * in any case.
      */
     public NonAnnotatedVariantsMongoReader(MongoOperations template, String collectionsVariantsName, String studyId) {
-        setName(ClassUtils.getShortName(NonAnnotatedVariantsMongoReader.class));
         if (studyId == null) {
             throw new IllegalArgumentException("NonAnnotatedVariantsMongoReader needs a non-null studyId " +
                     "(it can take a studyId or an empty string for reading every study)");
@@ -83,8 +85,14 @@ public class NonAnnotatedVariantsMongoReader extends AbstractItemCountingItemStr
     }
 
     @Override
-    protected VariantWrapper doRead() throws Exception {
-        DBObject dbObject = delegateReader.doRead();
+    public void open(ExecutionContext executionContext) throws ItemStreamException {
+        delegateReader.open(executionContext);
+    }
+
+    @Override
+    public VariantWrapper read()
+            throws Exception, UnexpectedInputException, ParseException, NonTransientResourceException {
+        DBObject dbObject = delegateReader.read();
         if (dbObject != null) {
             Variant variant = converter.convertToDataModelType(dbObject);
             return new VariantWrapper(variant);
@@ -94,12 +102,12 @@ public class NonAnnotatedVariantsMongoReader extends AbstractItemCountingItemStr
     }
 
     @Override
-    protected void doOpen() throws Exception {
-        delegateReader.doOpen();
+    public void update(ExecutionContext executionContext) throws ItemStreamException {
+        delegateReader.update(executionContext);
     }
 
     @Override
-    protected void doClose() throws Exception {
-        delegateReader.doClose();
+    public void close() throws ItemStreamException {
+        delegateReader.close();
     }
 }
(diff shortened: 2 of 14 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
```diff
@@ -31,39 +41,73 @@
...
-     *  If the studyId string is not empty, bring only non-annotated variants from that study. This parameter should
-     *  not be null in any case.
+     * @param studyId Can be the empty string, meaning to bring all non-annotated variants in the collection. If the
+     * studyId string is not empty, bring only non-annotated variants from that study. This parameter should not be null
+     * in any case.
      */
     public NonAnnotatedVariantsMongoReader(MongoOperations template, String collectionsVariantsName, String studyId) {
         if (studyId == null) {
             throw new IllegalArgumentException("NonAnnotatedVariantsMongoReader needs a non-null studyId " +
                     "(it can take a studyId or an empty string for reading every study)");
```
cyenyxe: I know we discussed it #100 , but I'm still not convinced about this interface.

What do you think of allowing the study ID to be null, and check that along with being empty before adding it as a condition? That would ensure that all cases of "unspecified study" behave the same way.

src/test/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReaderTest.java
```diff
@@ -103,12 +104,16 @@ private void checkNonAnnotatedVariantsRead(int expectedNonAnnotatedVariants, Str
...
-        DBObject variantMongoDocument;
-        while ((variantMongoDocument = mongoItemReader.read()) != null) {
+        VariantWrapper variantWrapper;
+        while ((variantWrapper = mongoItemReader.read()) != null) {
             itemCount++;
-            assertTrue(variantMongoDocument.containsField(VariantToDBObjectConverter.CHROMOSOME_FIELD));
-            assertTrue(variantMongoDocument.containsField(VariantToDBObjectConverter.START_FIELD));
-            assertFalse(variantMongoDocument.containsField(VariantToDBObjectConverter.ANNOTATION_FIELD));
+            assertFalse(variantWrapper.getChr().isEmpty());
+            assertFalse(variantWrapper.getStart() == 0);
```
cyenyxe: assertNotEquals?

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
```diff
@@ -31,39 +41,73 @@
...
+        String[] fields = {"chr", "start", "end", "ref", "alt"};
+        delegateReader.setFields(fields);
+
+        converter = new DBObjectToVariantConverter();
     }
 
     @PostConstruct
     @Override
     public void afterPropertiesSet() throws Exception {
-        super.afterPropertiesSet();
```
cyenyxe: Shouldn't we call the `super` method anyway?

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
```diff
@@ -31,39 +41,73 @@
  * {@link org.springframework.batch.item.data.MongoItemReader} is using
  * pagination and it is slow with large collections
  */
-public class NonAnnotatedVariantsMongoReader extends MongoDbCursorItemReader {
+public class NonAnnotatedVariantsMongoReader implements ItemStreamReader<VariantWrapper>, InitializingBean {
```
cyenyxe: Have you checked what state is stored in the context after making this an `ItemStreamReader`? I don't think it will be even possible to keep the cursor in a consistent state if an error occurs and the application is shut down.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
jmmut: no, we are not extending any class now, only implementing, and we already call `delegate.afterPropertiesSet()`

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
jmmut: there was not really any change, we implement the same interfaces as before, and calling the delegate instead of the (previously) extended class

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
```diff
@@ -31,39 +41,73 @@
...
+    }
+
+    @Override
+    public void open(ExecutionContext executionContext) throws ItemStreamException {
+        delegateReader.open(executionContext);
+    }
+
+    @Override
+    public VariantWrapper read()
+            throws Exception, UnexpectedInputException, ParseException, NonTransientResourceException {
```
poggio84: Looks like the last three are not necessary

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
poggio84: Looks to me a lot of convolution just to convert a DBObject into a VariantWrapper, I can see why we are doing it but I'm not sure if this is the best way. Anyway, I think that `AnnotationProcessor` and the test are no longer necessary then...

cyenyxe: The alternative we saw was to make `MongoDbCursorItemReader` do the conversion internally, but that could complicate the implementation more than wrapping it. Any suggestions welcome though.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/job/GenotypedVcfJobParametersValidator.java
```diff
@@ -52,6 +53,8 @@ private CompositeJobParametersValidator compositeJobParametersValidator(JobParam
 
         Boolean skipAnnotation = Boolean.valueOf(jobParameters.getString(JobParametersNames.ANNOTATION_SKIP));
         if (!skipAnnotation) {
+            jobParametersValidators.add(new InputStudyIdValidator());
```
cyenyxe: Why has this been taken out of the step, is it because it depends on the job instead of the step? Could a comment be added to clarify it?

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
poggio84: Yes I thought the same.. what about something similar to the LineMapper (or RowMapper) in the `FlatFileItemReader` some conversion/mapping logic that we can inject/set ???

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
jmmut: I kept them to fit the interface, but now I think about it, it doesn't really add much, I will remove them

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
jmmut: Specifically, what convolution are you referring to? The structure of this class is similar to a decorator pattern: we have the generic MongoDbCursorItemReader, and the AnnotatedVariantsMongoReader narrows the goal to read VariantWrappers, and for that it needs a specific query and a specific conversion (and the latter is already contained in the DBObjectToVariantConverter)

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/job/GenotypedVcfJobParametersValidator.java
jmmut: yes, I suggested to put the studyIdValidator as a required parameter to the stepValidator constructor (converting a logical dependency into a syntactical dependency) but with @jorizci we decided this was simpler. I'll add a comment. 

I still prefer my suggestion, it avoids logical programming errors without the need of comments for clarity

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
jorizci: I think it's more a delegate pattern, anyway the only other "solution" is to extend directly the `MongoDbCursorItemReader` and include a registration of a converter of type T.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/job/GenotypedVcfJobParametersValidator.java
jorizci: InputStudyId now has always a empty default value to avoid having three different states: null, empty, value. By having a empty value by default we reduce the problem so that if you only want to execute that step is not mandatory , but if you are using the job it is mandatory.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
jmmut: the validators and the InputParameters now take care of the null issue. the user can either not specify the study property (null) or leave it empty (""), and this class will receive an empty string in both cases. the variant load jobs use an extra validator so that it is not empty.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
jmmut: and the unneeded classes will be removed in a future PR I am incubating

src/main/java/uk/ac/ebi/eva/pipeline/configuration/readers/NonAnnotatedVariantsMongoReaderConfiguration.java
```diff
@@ -21,6 +21,7 @@
 import org.springframework.data.mongodb.core.MongoOperations;
 
 import uk.ac.ebi.eva.pipeline.io.readers.NonAnnotatedVariantsMongoReader;
+import uk.ac.ebi.eva.pipeline.parameters.AnnotationParameters;
```
cyenyxe: Unused import.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
cyenyxe: I still don't see a reason why providing null should throw an exception while being empty doesn't. I think both should be treated the same way (not throwing an exception but checked in line 66).

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/job/AggregatedVcfJobParametersValidator.java
```diff
@@ -50,6 +51,9 @@ private CompositeJobParametersValidator compositeJobParametersValidator(JobParam
 
         Boolean skipAnnotation = Boolean.valueOf(jobParameters.getString(JobParametersNames.ANNOTATION_SKIP));
         if (!skipAnnotation) {
+            // the next validator is required only in some jobs that use the annotation steps, the steps don't include it
```
cyenyxe: Please clarify why **this** job needs it. Applies to the other VCF processing job.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/NonAnnotatedVariantsMongoReader.java
jmmut: mmh, now I think about it again, I don't see the reason either. the change in the validators can take care of both the empty and null cases.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/job/AggregatedVcfJobParametersValidator.java
jmmut: changed usage

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/job/GenotypedVcfJobParametersValidator.java
```diff
@@ -53,11 +52,10 @@ private CompositeJobParametersValidator compositeJobParametersValidator(JobParam
 
         Boolean skipAnnotation = Boolean.valueOf(jobParameters.getString(JobParametersNames.ANNOTATION_SKIP));
         if (!skipAnnotation) {
-            // the next validator is required only in some jobs that use the annotation steps, the steps don't include it
-            jobParametersValidators.add(new InputStudyIdValidator());
+            boolean studyIdRequired = true;
 
             jobParametersValidators.add(new VepInputGeneratorStepParametersValidator());
-            jobParametersValidators.add(new VepAnnotationGeneratorStepParametersValidator());
+            jobParametersValidators.add(new VepAnnotationGeneratorStepParametersValidator(studyIdRequired));
```
jmmut: well, this PR won't go into develop until the feature/batch-annotation is finished, and VepAnnotationGeneratorStepParametersValidator is where the reader will be used

jmmut: but in fact, the annot load step validator does require that logic as well, and do it as well on the vcf_id parameter, so I'll add those changes

PR #110, opened 2017-04-03
EVA-659 Batch-friendly VEP run step

jmmut:
using #109 and #108, we can redo the annotation flow:

- annotation flow: --- **current PR** ---
    - GenerateAnnotationStep --- **current PR** ---
        - NonAnnotatedVariantsReader #109
        - VepAnnotationFileWriter #108
    - AnnotationLoaderStep
        - VariantAnnotationReader
        - VariantAnnotationWriter

other things done:
- there were a lot of classes removed, so I run a global imports optimization.
- VepProcess is opened lazily, because it fails if receives EOF without any variant (like failing on an empty list to annotate). This way, we only fire up VEP if there's at least one variant.
-  extract timeout to job parameter


src/main/java/uk/ac/ebi/eva/pipeline/configuration/writers/VepAnnotationFileWriterConfiguration.java
```diff
@@ -19,19 +19,23 @@
...
+public class VepAnnotationFileWriterConfiguration {
 
-    @Bean(VEP_INPUT_WRITER)
+    @Bean(VEP_ANNOTATION_WRITER)
     @StepScope
-    public ItemStreamWriter<VariantWrapper> vepInputFlatFileWriter(AnnotationParameters annotationParameters) {
-        return new VepInputFlatFileWriter(annotationParameters.getVepInput());
+    public ItemStreamWriter<VariantWrapper> vepAnnotationFileWriter(AnnotationParameters annotationParameters,
+            ChunkSizeParameters chunkSizeParameters) {
+        long timeoutInSeconds = 600L;   // TODO extract as a pipeline parameter
```
jorizci: If TODO then DO. It's not so much code to include it.

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
```diff
@@ -32,18 +32,25 @@
 
     private final VepProcess vepProcess;
 
-    public VepAnnotationFileWriter(Long timeoutInSeconds, Integer chunkSize, AnnotationParameters annotationParameters) {
+    private boolean vepProcessOpened = false;
```
jorizci: If you get the open logic from VepProces, something like vepProcess.isOpen() then you can remove the boolean flags / initialization

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
```diff
@@ -68,7 +75,9 @@ public void update(ExecutionContext executionContext) throws ItemStreamException
 
     @Override
     public void close() throws ItemStreamException {
-        vepProcess.close();
+        if (vepProcessOpened) {
```
jorizci: Don't remember, but I think VepProcess doesn't have problems with double close?

jmmut: yes it does, because it has to handle the error output when it closes the process (among other little things

jmmut: Oh, you meant that the VepProcess takes care of the double close, so this can be a direct call. yes, that's true, I'll change that back.

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepProcess.java
```diff
@@ -118,21 +137,28 @@ private void captureOutput(Process process, String vepOutput) {
...
     public void write(byte[] bytes) throws IOException {
-        if (process == null) {
+        if (!isOpen()) {
             throw new IllegalStateException("Process must be initialized (hint: call open() before write())");
         }
         processStandardInput.write(bytes);
     }
 
+    public boolean isOpen() {
+        return process != null;
```
cyenyxe: A call to [Process.isAlive()](https://docs.oracle.com/javase/8/docs/api/java/lang/Process.html#isAlive--) could be also relevant, as the documentation states:

> Returns:
true if the subprocess represented by this Process object has not yet terminated.

So if the process has died for some reason between writes, that is the method that would catch it. That implies changing a bit the exception message or showing 2 different messages depending on the situation.

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepProcess.java
```diff
@@ -118,21 +137,28 @@ private void captureOutput(Process process, String vepOutput) {
     }
 
     public void write(byte[] bytes) throws IOException {
```
cyenyxe: Could a class be created to test the workflow for the methods (write, flush, close...) that check if a process is open?

src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/EmptyVepOutputDecider.java
```diff
@@ -26,36 +26,37 @@
...
 
     @Override
     public FlowExecutionStatus decide(JobExecution jobExecution, StepExecution stepExecution) {
-        String vepInput = getVepInput(jobExecution);
+        String vepOutput = getVepOutput(jobExecution);
 
-        if (getFileSize(vepInput) <= 0) {
-            logger.info("File {} is empty so following steps will not run", vepInput);
+        if (getFileSize(vepOutput) <= 0) {
+            logger.info("File {} is empty so following steps will not run", vepOutput);
```
cyenyxe: "subsequent" instead of "following" maybe? @tomdcsmith help!

src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/EmptyVepOutputDecider.java
```diff
@@ -64,10 +65,14 @@ private String getVepInput(JobExecution jobExecution) {
...
-            fileSize = Files.size(Paths.get(file));
-        } catch (IOException e) {
-            throw new RuntimeException("File {} is not readable", e);
+        if (!new File(file).exists()) {
+            fileSize = 0;
+        } else {
+            try {
+                fileSize = Files.size(Paths.get(file));
+            } catch (IOException e) {
+                throw new RuntimeException("File " + file + " is not readable", e);
```
cyenyxe: Would it be that inconvenient to wrap the IOException with another (checked) IOException? If it is not readable, we want to make sure the issue is properly reported, instead of the exception re-thrown indefinitely.

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/GenerateVepAnnotationStep.java
```diff
@@ -17,38 +17,71 @@
...
+import static uk.ac.ebi.eva.pipeline.configuration.BeanNames.NON_ANNOTATED_VARIANTS_READER;
+import static uk.ac.ebi.eva.pipeline.configuration.BeanNames.VEP_ANNOTATION_WRITER;
 
 /**
- * Configuration class that inject a step created with the tasklet {@link VepAnnotationGeneratorStep}
+ * This step creates a file with variant annotations.
+ * <p>
+ * input: mongo collection with the variants. Only non-annotated variants will be retrieved.
+ * <p>
+ * output: file with the list of annotated variants, in a format written by VEP, readable with ({@link AnnotationFlatFileReader})
```
cyenyxe: Please care capitalization.

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/GenerateVepAnnotationStep.java
```diff
@@ -17,38 +17,71 @@
...
+ * <p>
+ * TODO:
```
cyenyxe: Do these TODO messages still apply? I mean, ticket https://github.com/EBIvariation/eva-v2/pull/22 goes back to June 2016...

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/job/AggregatedVcfJobParametersValidator.java
```diff
@@ -52,8 +51,7 @@ private CompositeJobParametersValidator compositeJobParametersValidator(JobParam
         if (!skipAnnotation) {
             boolean studyIdRequired = true;
 
-            jobParametersValidators.add(new VepInputGeneratorStepParametersValidator());
-            jobParametersValidators.add(new VepAnnotationGeneratorStepParametersValidator(studyIdRequired));
+            jobParametersValidators.add(new GenerateVepAnnotationStepParametersValidator(studyIdRequired));
             jobParametersValidators.add(new AnnotationLoaderStepParametersValidator(studyIdRequired));
             jobParametersValidators.add(new AnnotationMetadataStepParametersValidator());
```
cyenyxe: Inconsistent naming across these 3 steps, please make them more homogeneous taking other steps into account.

src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/EmptyVepOutputDecider.java
jmmut: we can not raise exceptions, the overriden `public FlowExecutionStatus decide(JobExecution jobExecution, StepExecution stepExecution)` doesn't allow it.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/job/AggregatedVcfJobParametersValidator.java
jmmut: I would leave that for the next PR, about renamings and package movings.

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepProcess.java
jmmut: not sure about adding this. if it died suddenly, trying to use it would result in an exception raised, which is what we want. 

On top of that, mixing `process != null` and `Process.isAlive()` would mix concerns on API usage and runtime validity. Using that hypothetical method would be also verbose because of that mixing

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/GenerateVepAnnotationStep.java
jmmut: no, we are already writing gzipped without overwriting

src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/EmptyVepOutputDecider.java
tomdcsmith: I agree that "subsequent" fits better.

src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/EmptyVepOutputDecider.java
```diff
@@ -26,36 +26,37 @@
...
     public static final String CONTINUE_FLOW = "CONTINUE_FLOW";
 
     @Override
     public FlowExecutionStatus decide(JobExecution jobExecution, StepExecution stepExecution) {
-        String vepInput = getVepInput(jobExecution);
+        String vepOutput = getVepOutput(jobExecution);
 
-        if (getFileSize(vepInput) <= 0) {
-            logger.info("File {} is empty so following steps will not run", vepInput);
+        if (getFileSize(vepOutput) <= 0) {
```
cyenyxe: In which case could this be smaller than zero?

src/test/java/uk/ac/ebi/eva/pipeline/io/VepProcessTest.java
```diff
@@ -0,0 +1,109 @@
...
+    public void testWorkflowWriteWithoutOpening() throws Exception {
+        DBObjectToVariantConverter converter = new DBObjectToVariantConverter();
+        VariantWrapper variantWrapper = new VariantWrapper(
+                converter.convertToDataModelType(constructDbObject(VariantData.getVariantWithAnnotation())));
+        annotationParameters.setVepPath(getResource("/mockvep_writeToFile_delayed.pl").getAbsolutePath());
+
+        VepProcess vepAnnotationFileWriter = new VepProcess(annotationParameters, CHUNK_SIZE, VEP_TIMEOUT);
+
+        exception.expect(IllegalStateException.class);
+        exception.expectMessage(Matchers.contains(HINT_CALL_OPEN));
```
cyenyxe: Checking message contents not externalized to configuration files is quite risky: changing a single character will break the test. If the test case set up is simplified (as in the previous comment) it should be enough to just expect an IllegalStateException for the whole method.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/steps/GenerateVepAnnotationStepTest.java
```diff
@@ -92,10 +103,10 @@ public void shouldGenerateVepAnnotations() throws Exception {
...
 
         // And VEP output should exist and annotations should be in the file
-        File vepOutputFile = new File(URLHelper.resolveVepOutput(vepOutputFolder.getAbsolutePath(), STUDY_ID, FILE_ID));
-
-        assertTrue(vepOutputFile.exists());
-        assertEquals(537, JobTestUtils.getLines(new GZIPInputStream(new FileInputStream(vepOutputFile))));
+        assertTrue(vepOutput.exists());
+        int EXTRA_ANNOTATIONS = 1;
+        assertEquals(300 + EXTRA_ANNOTATIONS,
+                JobTestUtils.getLines(new GZIPInputStream(new FileInputStream(vepOutput))));
```
cyenyxe: Could you please explain why has this number changed so drastically?

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/VepInputFlatFileWriterTest.java
```diff
@@ -1,62 +0,0 @@
...
- * input: a List of VariantWrapper to each call of `.write()`
- * output: a line in a File for each VariantWrapper
- */
-public class VepInputFlatFileWriterTest {
-
-    @Rule
-    public PipelineTemporaryFolderRule temporaryFolderRule = new PipelineTemporaryFolderRule();
-
-    @Test
-    public void shouldWriteAllFieldsToFile() throws Exception {
```
jmmut: This is Vep* **Input** *FlatFileWriterTest. It is similar to VepAnnotationFileWriterTest, but this new one is meant to replace the previous annotation tasklet with variantWrappers as input. I mean, the structure is quite different and there is no easy 1 to 1 mapping of code and tests

src/test/java/uk/ac/ebi/eva/pipeline/io/VepProcessTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+        File annotationFolder = temporaryFolder.newFolder();
+        annotationParameters.setOutputDirAnnotation(annotationFolder.getAbsolutePath());
+    }
+
+    @Test
+    public void testWorkflowWriteWithoutOpening() throws Exception {
+        DBObjectToVariantConverter converter = new DBObjectToVariantConverter();
+        VariantWrapper variantWrapper = new VariantWrapper(
+                converter.convertToDataModelType(constructDbObject(VariantData.getVariantWithAnnotation())));
```
jmmut: It has only one constructor that takes a Variant. I will add another constructor that takes only the required parameters, scaffolds a Variant, and adjusts its coordinates to ensembl format. Let me know if it makes sense to change it completely to a "EnsembleCoordinates" and have only the split fields that the getters provide.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/steps/GenerateVepAnnotationStepTest.java
jmmut: the VepAnnotationFileWriterTest and VepProcessTest could not work with the existing mockvep.pl, so I added two more versions of that script. After that, for the other integration tests I had to modify the old mockvep, and now this test can work with both scripts. 

The old script now takes all data from standard input, ignores it, and writes a hardcoded (correct) annotation. the new script copies the input into the output in batches, adding "annotated" to each line (not actual annotation), and adds 1 extra line (vep can do this, and we need to check that we support that behaviour). There's a third script similar to the second one that has sleeps for the timeout checks

src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/EmptyVepOutputDecider.java
jmmut: from `Files.size()`: 
> The size of files that are not regular files is implementation specific and therefore unspecified.

Indeed it's actually unlikely that we get a smaller than zero size, but I think that check does no harm being there

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/VepAnnotationFileWriter.java
```diff
@@ -32,18 +33,22 @@
 
     private final VepProcess vepProcess;
 
-    public VepAnnotationFileWriter(Long timeoutInSeconds, Integer chunkSize, AnnotationParameters annotationParameters) {
+    public VepAnnotationFileWriter(AnnotationParameters annotationParameters, Integer chunkSize, Long timeoutInSeconds) {
```
jmmut: mmh, after changing this, I saw that making it primitive, becomes required in other steps like annotation load, or annotation metadata.

We are checking that the parameter is provided in the generateAnnotationStep parameters validator, so I would leave it as Long.

src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/EmptyVepOutputDecider.java
jorizci: Or if we could just do the opposite...   >0 -> continue else stop.

jmmut:
## Update

given that we needed to sort the query for the NonAnnotatedVariantsMongoReader, and the problems with sorting are not assumable at this point, we don't skip any chunk if the step failed previously. We rename the failed annotation file (vepOutput) and start again the step.

src/main/java/uk/ac/ebi/eva/pipeline/io/VepProcess.java
```diff
@@ -0,0 +1,341 @@
...
+                    logger.debug("Extending the timeout, as the process wrote more lines (it's still active)");
+                }
+            } while (processWroteDuringWait && !finished);
+        } catch (InterruptedException e) {
+            throw new ItemStreamException(e);
+        }
+
+        if (!finished) {
+            String timeoutReachedMessage = "VEP has been idle for more than the timeout (" + timeoutInSeconds
+                    + " seconds). Killed the process.";
```
cyenyxe: `The process has been killed.`

src/main/java/uk/ac/ebi/eva/pipeline/io/VepProcess.java
```diff
@@ -0,0 +1,341 @@
...
+            logger.error(timeoutReachedMessage);
```
cyenyxe: The message should be logged **after** the process is actually destroyed, otherwise it will be displayed even if it could not be destroyed successfully.

src/main/java/uk/ac/ebi/eva/pipeline/io/VepProcess.java
```diff
@@ -0,0 +1,341 @@
...
+
+    private Process process;
+
+    private OutputStream processStandardInput;
+
+    private Thread outputCapturer;
+
+    private AtomicBoolean writingOk;
+
+    private static final boolean APPEND = true;
```
cyenyxe: Move static final attributes to the beginning of the declarations block.

src/main/java/uk/ac/ebi/eva/pipeline/io/VepProcess.java
```diff
@@ -0,0 +1,341 @@
...
+                logger.info("Renamed VEP output to " + backupVepOutput + " as backup because VEP failed");
+            }
+            throw new ItemStreamException("Error while running VEP (exit status " + exitValue + "). See "
+                    + errorLog + " for the errors description from VEP.");
+        }
+    }
+
+    private void checkOutputWritingStatus() {
+        try {
+            outputCapturer.join(timeoutInSeconds * CONVERT_SECONDS_TO_MILLISECONDS);
```
cyenyxe: Maybe a bit extreme? `timeoutInSeconds * 1000` would be clearly in milliseconds.

src/main/java/uk/ac/ebi/eva/pipeline/io/VepProcess.java
```diff
@@ -0,0 +1,341 @@
...
+        } catch (InterruptedException e) {
+            throw new ItemStreamException("Interrupted while waiting for the VEP output writer thread to finish. ", e);
+        }
+        if (outputCapturer.isAlive()) {
+            outputCapturer.interrupt();
+            throw new ItemStreamException("Reached the timeout (" + timeoutInSeconds
+                    + " seconds) while waiting for VEP output writing to finish. Killed the thread.");
```
cyenyxe: `The thread has been killed`

src/main/java/uk/ac/ebi/eva/pipeline/io/VepProcess.java
```diff
@@ -0,0 +1,341 @@
...
+            writer.write('\n');
+            writtenLines++;
+
+            lastLine = line;
+            line = getNextLine(reader, skipComments);
+        }
+
+        writer.flush();
+        outputIdleSince.set(System.currentTimeMillis());
+        logCoordinates(lastLine, writtenLines);
```
cyenyxe: Can `line` be null after the first `getNextLine` call? If that is the case, then `lastLine` will also be, the while loop won't execute and `logCoordinates` will fail when calling `isComment`.

src/main/java/uk/ac/ebi/eva/pipeline/io/VepProcess.java
```diff
@@ -0,0 +1,341 @@
...
+            } else {
+                Scanner scanner = new Scanner(line);
+                logger.trace("VEP wrote {} more lines, last one was {}", chunkSize, scanner.next());
+            }
+        }
+    }
+
+    /**
+     * read all the inputStream and write it into the outputStream. This method blocks until input data is available,
+     * the end of the stream is detected, or an exception is thrown.
```
cyenyxe: `Read the whole input stream and write it into the output stream`

src/main/java/uk/ac/ebi/eva/pipeline/jobs/deciders/EmptyVepOutputDecider.java
cyenyxe: The way @jorizci suggests is well defined behaviour so it could be safer.

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/GenerateVepAnnotationStep.java
```diff
@@ -17,38 +17,65 @@
...
+    @Qualifier(NON_ANNOTATED_VARIANTS_READER)
+    private ItemStreamReader<VariantWrapper> nonAnnotatedVariantsReader;
+
+    @Autowired
+    @Qualifier(VEP_ANNOTATION_WRITER)
+    private ItemWriter<VariantWrapper> vepAnnotationWriter;
 
     @Bean(GENERATE_VEP_ANNOTATION_STEP)
-    public TaskletStep generateVepAnnotationStep(StepBuilderFactory stepBuilderFactory, JobOptions jobOptions) {
+    public Step generateVepInputStep(StepBuilderFactory stepBuilderFactory, JobOptions jobOptions,
```
cyenyxe: Why not using the old name? If there is conflict `generateVepOutputStep` could be used. But this step is definitely **not** generating VEP input.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/steps/GenerateVepAnnotationStepTest.java
```diff
@@ -0,0 +1,188 @@
...
+    @Rule
+    public TemporaryMongoRule mongoRule = new TemporaryMongoRule();
+
+    @Rule
+    public PipelineTemporaryFolderRule temporaryFolderRule = new PipelineTemporaryFolderRule();
+
+    @Autowired
+    private JobLauncherTestUtils jobLauncherTestUtils;
+
+    private static final int EXTRA_ANNOTATIONS = 1;
```
cyenyxe: Move with the rest of static final declarations.

src/main/java/uk/ac/ebi/eva/pipeline/io/VepProcess.java
```diff
@@ -0,0 +1,341 @@
...
+            try {
+                connectStreams(process.getErrorStream(), new FileOutputStream(errorLog));
+            } catch (IOException e) {
+                throw new ItemStreamException("VEP exited with code " + exitValue
+                        + " but the file to dump the errors could not be created: " + errorLog,
+                        e);
+            }
+            boolean renamed = new File(annotationParameters.getVepOutput()).renameTo(new File(backupVepOutput));
+            if (renamed) {
+                logger.info("Renamed VEP output to " + backupVepOutput + " as backup because VEP failed");
```
jorizci: 2 -> 1 -> 3

src/test/resources/mockvep_writeToFile_error.pl
```diff
@@ -0,0 +1,49 @@
+use warnings;
```
jorizci: Please document the peculiarities of this scrip.

src/test/resources/mockvep_writeToFile_delayed.pl
```diff
@@ -38,7 +38,7 @@
     print $fileHandle $bufferLine;
 }
 
+sleep 3;
```
jorizci: Please document the peculiarities of this scrip.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/AnnotationParameters.java
```diff
@@ -63,6 +63,9 @@
     @Value(PARAMETER + JobParametersNames.APP_VEP_NUMFORKS + END)
     private String vepNumForks;
 
+    @Value(PARAMETER + JobParametersNames.APP_VEP_TIMEOUT + END)
+    private Long timeout;
```
jmmut: I launched it several times in the terminal, I think it's not a problem anymore (chunk size is an Integer as well). I'll change the others to number types

src/main/resources/application.properties
```diff
@@ -32,3 +32,5 @@ logging.level.uk.ac.ebi.eva=DEBUG
 logging.level.org.opencb.opencga=DEBUG
 logging.level.org.springframework=INFO
 
+app.vep.timeout=600
```
jmmut: good catch

src/main/java/uk/ac/ebi/eva/pipeline/io/VepProcess.java
jmmut: It won't, because VEP fails if it gets no input, and we take care of that in VepAnnotationFileWriter. The only way I can imagine that, is VEP getting a variant and writing nothing. but well, I can add a check for that:

    logger.warn("VEP didn't write anything, this might be a symptom of a previous error");

src/main/java/uk/ac/ebi/eva/pipeline/listeners/StepProgressListener.java
```diff
@@ -36,22 +36,24 @@ public void beforeChunk(ChunkContext context) {
     @Override
     public void afterChunk(ChunkContext context) {
 
-        long estimatedTotalNumberOfLines = (long)context.getStepContext().getStepExecutionContext()
+        Object lines = context.getStepContext().getStepExecutionContext()
```
jmmut: nope, execution context is a map<String, Object>, the casting is still required (even for Object to Long).

I could do the next to join the retrieval, default value and casting: 
```
long estimatedTotalNumberOfLines = (long) context.getStepContext().getStepExecutionContext()
    .getOrDefault(ExecutionContextParametersNames.NUMBER_OF_LINES, 0L);
```

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/GenerateVepAnnotationStep.java
jmmut: Every renaming storm lets some sand in your pants. reverting...

PR #114, opened 2017-04-26
EVA-719 Monitor progress of variant annotation load

poggio84:
- _VcfNumberOfLinesEstimator_ renamed to a more generic _FileWithHeaderNumberOfLinesEstimator_
- Listener added to Annotation loader step
- Test updated including a test for annotation file


src/main/java/uk/ac/ebi/eva/utils/FileWithHeaderNumberOfLinesEstimator.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.Scanner;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Estimate the number of lines in a file.
+ */
+public class FileWithHeaderNumberOfLinesEstimator {
+    private static final Logger logger = LoggerFactory.getLogger(FileWithHeaderNumberOfLinesEstimator.class);
+
+    private static final int NUMBER_OF_LINES = 150;
```
jmmut: After some discussion, we agreed to increase this number to 10K to get more precision, but then, if the file is shorter than that, it is reasonable to output the actual line number, as we know that at some point

src/main/java/uk/ac/ebi/eva/utils/FileWithHeaderNumberOfLinesEstimator.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private boolean skipEstimation(long linesReadInBody) {
+        return linesReadInBody < MAX_NUMBER_OF_LINES;
+    }
+
+    /**
+     * @return lines in the body. It will be min(MAX_NUMBER_OF_LINES, actualLinesInTheBody)
+     */
+    private long copyHeadAndBody(String filePath) {
```
cyenyxe: Agreed offline to create 2 functions, one to copy the head and another to copy the body, which will call a private one with the common behavior.

src/main/java/uk/ac/ebi/eva/utils/FileWithHeaderNumberOfLinesEstimator.java
```diff
@@ -0,0 +1,133 @@
...
+        return new OutputStreamWriter(new GZIPOutputStream(new FileOutputStream(file)), "UTF-8");
+    }
+
+    private long getEstimatedTotalNumberOfLines(String filePath) {
+        long fileSize = new File(filePath).length();
+        long singleLineSize = bodyFile.length() / MAX_NUMBER_OF_LINES;
+        long headFileSize = headFile.length();
+
+        boolean headDeleted = headFile.delete();
+        boolean bodyDeleted = bodyFile.delete();
```
cyenyxe: [File.deleteOnExit()](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#deleteOnExit()) instead?

src/main/java/uk/ac/ebi/eva/utils/FileWithHeaderNumberOfLinesEstimator.java
jmmut: re agreed offline to let this as-is

PR #115, opened 2017-05-02
EVA-683 Detect when VEP hangs

jmmut:
It seems that when VEP hangs (due to the unnoticed death of some of its inner threads probably), we continue writing into the standard input (stdin) of VEP, but it won't write anything more, nor will consume more data from its stdin, so eventually the buffers get filled and it turns out that in the OutputStream class the `write` method can also block (just as `read`), but doesn't provide any "write with timeout" method.

So, what we have to do is keep track of the time we spent writing, and stop the step if the timeout is reached.

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/io/VepProcess.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/io/VepProcess.java
@@ -33,7 +33,12 @@ import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.util.Arrays;
 import java.util.Scanner;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/io/VepProcess.java
```diff
@@ -168,7 +175,25 @@ public void flush() throws IOException {
...
             throw new IllegalStateException("Process must be initialized (hint: call open() before flush())");
         }
-        processStandardInput.flush();
+        tryWithTimeout(() -> {
+            processStandardInput.flush();
+            return null;
+        });
+    }
+
+    private void tryWithTimeout(Callable<Void> callable) {
```
cyenyxe: This method is placed a bit arbitrarily: it is not after its first usage nor is at the beginning of a set of private methods.

PR #117, opened 2017-05-09
EVA-741 The parameters from completed executions should not be reused

jmmut:
The second time a job is launched, it should only reuse parameters (that can be overwritten with the current parameters) if the job failed. Otherwise, only the new parameters should be used. 

A solution is to implement our own JobParameters incrementer, because it will be used here https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherCommandLineRunner.java#L158 (only if the job is not stopped or failed)

### DIFF
--- a/src/main/java/uk/ac/ebi/eva/pipeline/parameters/NewJobIncrementer.java
+++ b/src/main/java/uk/ac/ebi/eva/pipeline/parameters/NewJobIncrementer.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2015-2017 EMBL - European Bioinformatics Institute
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.ac.ebi.eva.pipeline.parameters;
+
+import org.springframework.batch.core.JobParameters;
+import org.springframework.batch.core.JobParametersBuilder;
+import org.springframework.batch.core.JobParametersIncrementer;
+
+/**
+ * Incrementer that does not reuse parameters from previous completed executions.
+ */
+public class NewJobIncrementer implements JobParametersIncrementer {
+
+    private static String RUN_ID_KEY = "run.id";
+
+    private String key = RUN_ID_KEY;
+
+    /**
+     * The name of the run id in the job parameters.  Defaults to "run.id".
+     *
+     * @param key the key to set
+     */
+    public void setKey(String key) {
+        this.key = key;
+    }
+
+    /**
+     * Increment the run.id parameter (starting with 1), and ignore the other parameters.
+     */
+    @Override
+    public JobParameters getNext(JobParameters parameters) {
+        JobParameters params = (parameters == null) ? new JobParameters() : parameters;
+
+        long id = params.getLong(key, 0L) + 1;
+        return new JobParametersBuilder().addLong(key, id).toJobParameters();
+    }
+}
--- a/src/test/java/uk/ac/ebi/eva/pipeline/jobs/GenotypedVcfJobTest.java
+++ b/src/test/java/uk/ac/ebi/eva/pipeline/jobs/GenotypedVcfJobTest.java
@@ -36,11 +36,13 @@ import uk.ac.ebi.eva.test.configuration.BatchTestConfiguration;
 import uk.ac.ebi.eva.test.rules.PipelineTemporaryFolderRule;
 import uk.ac.ebi.eva.test.rules.TemporaryMongoRule;
 import uk.ac.ebi.eva.test.utils.GenotypedVcfJobTestUtils;
+import uk.ac.ebi.eva.test.utils.JobTestUtils;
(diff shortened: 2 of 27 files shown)


src/main/java/uk/ac/ebi/eva/pipeline/parameters/NewJobIncrementer.java
```diff
@@ -0,0 +1,50 @@
...
+    private static String RUN_ID_KEY = "run.id";
+
+    private String key = RUN_ID_KEY;
+
+    /**
+     * The name of the run id in the job parameters.  Defaults to "run.id".
+     *
+     * @param key the key to set
+     */
+    public void setKey(String key) {
```
cyenyxe: This is not part of the `JobParametersIncrementer` interface, maybe it can be removed?

src/test/java/uk/ac/ebi/eva/pipeline/runner/EvaPipelineJobLauncherCommandLineRunnerTest.java
```diff
@@ -168,6 +171,63 @@ public void genotypedVcfJobTest() throws JobExecutionException, IOException, URI
...
+
+        assertFalse(jobExplorer.getJobInstances(GENOTYPED_VCF_JOB, 0, 1).isEmpty());
+        JobExecution jobExecution = jobExplorer.getJobExecution(jobExplorer.getJobInstances(GENOTYPED_VCF_JOB, 0, 1)
+                .get(0).getInstanceId());
+
+        assertEquals(ExitStatus.COMPLETED, jobExecution.getExitStatus());
+        assertEquals(BatchStatus.COMPLETED, jobExecution.getStatus());
+
+        assertEquals(ExitStatus.COMPLETED, jobExecution.getExitStatus());
+        assertEquals(BatchStatus.COMPLETED, jobExecution.getStatus());
```
cyenyxe: Duplicated.

src/test/java/uk/ac/ebi/eva/pipeline/runner/EvaPipelineJobLauncherCommandLineRunnerTest.java
```diff
@@ -168,6 +171,63 @@ public void genotypedVcfJobTest() throws JobExecutionException, IOException, URI
...
+                .statisticsSkip(true);
+
+        String[] commandLineWithoutChunksize = evaCommandLineBuilderWithoutChunksize
+                .inputVcfId("second_" + GenotypedVcfJobTestUtils.INPUT_VCF_ID)
+                .build();
+
+        String[] completeCommandLine = evaCommandLineBuilderWithoutChunksize
+                .inputVcfId("first_" + GenotypedVcfJobTestUtils.INPUT_VCF_ID)
+                .chunksize("100")
+                .build();
```
cyenyxe: If this is the first one, could it be declared first? Otherwise it looks like an oversight that could involve weird behavior in the test.

src/main/java/uk/ac/ebi/eva/pipeline/parameters/NewJobIncrementer.java
jorizci: That is for the case that you want to override the used "key" for the incrementer.

src/test/java/uk/ac/ebi/eva/pipeline/runner/EvaPipelineJobLauncherCommandLineRunnerTest.java
jmmut: no it can't. the builder is a thin wrapper around a map, but only with `put` methods. I'm not even sure it would be easy to copy it.

src/test/java/uk/ac/ebi/eva/pipeline/runner/EvaPipelineJobLauncherCommandLineRunnerTest.java
```diff
@@ -139,32 +145,81 @@ public void genotypedVcfJobTest() throws JobExecutionException, IOException, URI
...
 
-        assertEquals(EvaPipelineJobLauncherCommandLineRunner.EXIT_WITHOUT_ERRORS,
-                evaPipelineJobLauncherCommandLineRunner.getExitCode());
+        assertEquals(EXIT_WITHOUT_ERRORS, evaPipelineJobLauncherCommandLineRunner.getExitCode());
 
-        assertFalse(jobExplorer.getJobInstances(GENOTYPED_VCF_JOB, 0, 1).isEmpty());
-        JobExecution jobExecution = jobExplorer.getJobExecution(jobExplorer.getJobInstances(GENOTYPED_VCF_JOB, 0, 1)
-                .get(0).getInstanceId());
+        JobExecution jobExecution = getLastJobExecution(GENOTYPED_VCF_JOB);
+        assertCompleted(jobExecution);
```
cyenyxe: This method could be in a utility class, because many classes are performing these checks (for completed, failed or whatever combination). Please make it consistent across the whole codebase, either using the method everywhere or going back to the 2 asserts in this class.

src/test/java/uk/ac/ebi/eva/pipeline/jobs/GenotypedVcfJobTest.java
```diff
@@ -163,7 +164,6 @@ public void aggregationIsNotAllowed() throws Exception {
                 .toJobParameters();
         JobExecution jobExecution = jobLauncherTestUtils.launchJob(jobParameters);
 
-        assertEquals(ExitStatus.FAILED, jobExecution.getExitStatus());
-        assertEquals(BatchStatus.FAILED, jobExecution.getStatus());
+        assertFailed(jobExecution);
```
cyenyxe: Be consistent in the way this is called: right above the `JobTestUtils` class is referenced.

PR #124, opened 2017-07-18
EVA-789 More resilient VEP annotation flow

jmmut:
Restructuring the VEP job. Changed from "create annotations file, when complete, then load" to "create annotation batch and load, then get next annotation batch". In case of failure, the successful annotations will already be in the DB.

The annotation job structure introduced by this PR is:
- annotationJob:
  - annotationFlow:
    - generateVepAnnotationStep (chunksize 1)
      - VariantsMongoReader: outputs `List<EnsemblVariant>`
      - AnnotationCompositeProcessor: takes `List<EnsemblVariant>`, outputs `List<Annotation>`
        - VepAnnotationProcessor: takes `List<EnsemblVariant>`, outputs `List<String>`
        - AnnotationParserProcessor: takes `List<String>`, outputs `List<Annotation>`
      - AnnotationCompositeWriter: takes `List<Annotation>`
        - AnnotationInVariantMongoWriter
        - AnnotationMongoWriter
    - annotationMetadataStep


src/main/java/uk/ac/ebi/eva/pipeline/configuration/jobs/flows/AnnotationFlowConfiguration.java
```diff
@@ -44,32 +43,22 @@
...
-    @Autowired
     @Qualifier(LOAD_ANNOTATION_METADATA_STEP)
     private Step annotationMetadataStep;
 
     @Bean(VEP_ANNOTATION_FLOW)
     public Flow vepAnnotationFlow() {
-        EmptyVepOutputDecider emptyVepOutputDecider = new EmptyVepOutputDecider();
-
         return new FlowBuilder<Flow>(VEP_ANNOTATION_FLOW)
                 .start(generateVepAnnotationStep)
```
jorizci: Have you checked that in the edge case that no variant is left to annotate this will work?

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VariantsMongoReader.java
```diff
@@ -99,16 +108,36 @@ protected void doOpen() throws Exception {
     }
 
     @Override
-    protected EnsemblVariant doRead() throws Exception {
-        DBObject dbObject = delegateReader.doRead();
-        if (dbObject != null) {
-            SimplifiedVariant variant = converter.read(SimplifiedVariant.class, dbObject);
-            return buildVariantWrapper(variant);
+    protected List<EnsemblVariant> doRead() throws Exception {
+        if (moreElementsAvailable) {
```
jorizci: I think that this function should be simplified and avoid the flag variable. Also the read could probably be done with a batch size and avoid the roundtrips for each element and still use the spring converter.

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/AnnotationMongoWriter.java
```diff
@@ -86,10 +86,12 @@ public AnnotationMongoWriter(MongoOperations mongoOperations, String collection)
     }
 
     @Override
-    public void write(List<? extends Annotation> annotations) throws Exception {
+    public void write(List<? extends List<Annotation>> annotations) throws Exception {
         BulkOperations bulk = mongoOperations.bulkOps(BulkOperations.BulkMode.UNORDERED, collection);
```
jorizci: Not sure of this, originally this was sound but if you only initialize once, I don't know if we can expect the underlying implementation to clean the elements in the bulk after each execute operation.

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/AnnotationMongoWriter.java
```diff
@@ -86,10 +86,12 @@ public AnnotationMongoWriter(MongoOperations mongoOperations, String collection)
...
-        prepareBulk(annotations, bulk);
-        bulk.execute();
+        for (List<Annotation> annotationList : annotations) {
+            prepareBulk(annotationList, bulk);
+            bulk.execute();
```
jorizci: Maybe we could join all the annotation lists? Or do you think that it will be really heavy?

src/main/java/uk/ac/ebi/eva/pipeline/configuration/jobs/flows/AnnotationFlowConfiguration.java
jmmut: yes, the reader will return null on the first batch, and the processor and writer won't be called

src/main/java/uk/ac/ebi/eva/pipeline/io/writers/AnnotationMongoWriter.java
jmmut: one bulk per sublist seems better

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VariantsMongoReader.java
```diff
@@ -46,25 +52,28 @@
...
 
-    private static final String STUDY_KEY = VariantDocument.FILES_FIELD + "." + VariantSourceEntryMongo.STUDYID_FIELD;
+    private Integer chunkSize;
 
     /**
-     * @param studyId Can be the empty string or null, meaning to bring all non-annotated variants in the collection.
-     * If the studyId string is not empty, bring only non-annotated variants from that study.
+     * @param studyId Can be the empty string or null, meaning to bring all non-annotated variants in the collection. If
+     * the studyId string is not empty, bring only non-annotated variants from that study.
      * @param excludeAnnotated bring only non-annotated variants.
```
pabarcgar: The param chucksize is not in this javadoc

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/processors/AnnotationParserProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public AnnotationParserProcessor(AnnotationParameters annotationParameters) {
+        annotationLineMapper = new AnnotationLineMapper(annotationParameters.getVepVersion(),
+                                                        annotationParameters.getVepCacheVersion());
+    }
+
+    @Override
+    public List<Annotation> process(List<String> ensemblVariants) throws Exception {
+        List<Annotation> annotations = new ArrayList<>();
+        for (String ensemblVariant : ensemblVariants) {
```
pabarcgar: This is not bad, but it seems that this loop could be replaced by something like:
ensemblVariants.stream().map(v -> annotationLineMapper.mapLine(v, ....)).collect(Collectors.toList());

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VariantsMongoReader.java
```diff
@@ -94,21 +104,33 @@ public void afterPropertiesSet() throws Exception {
...
 
+    private List<EnsemblVariant> fillBatch(DBObject dbObject) throws Exception {
+        List<EnsemblVariant> variants = new ArrayList<>();
+        while (dbObject != null) {
+            SimplifiedVariant variant = converter.read(SimplifiedVariant.class, dbObject);
+            variants.add(buildVariantWrapper(variant));
+            if (variants.size() == chunkSize) {
+                break;
+            }
+            dbObject = delegateReader.doRead();
```
pabarcgar: For me is confusing that the read() method is reading the first line and delegating the reading of the rest of the lines in the chunk to the fillBatch() one. The first time I read this class I thought it was returning a list with just one variant, and just in a second reading I noticed the real workflow. 

Can those two methods be refactored so is more clear that a chunk is being read and not just one element?

src/main/java/uk/ac/ebi/eva/pipeline/jobs/steps/processors/VepAnnotationProcessor.java
```diff
@@ -13,58 +13,72 @@
...
-            logger.trace("VEP has received {} variants from {}:{} to {}:{}", variantWrappers.size(),
-                    first.getChr(), first.getStart(), last.getChr(), last.getStart());
-        }
+        logBatch(ensemblVariants);
 
         vepProcess.flush();
         vepProcess.close();
+        writer.close();
+
+        String[] lines = writer.getBuffer().toString().split("\n");
```
pabarcgar: Why if instead passing an StringWriter to the VepProcess, and then extracting the lines from the writer to a list, we pass a list to the VepProcess? 

Right now we are writing the lines to the writer, then to an array, and finally to a list. If the batch is big enough, doing that change could save some memory ...

jmmut: this was a "legacy design" issue. Changing it to what you say implies redoing all the writing part in that class. the current solution implied to extract the writer and add the line you point. I'm ok changing the interface of VepProcess, but it would mean imposing tighter interface constraints in the class.

src/test/java/uk/ac/ebi/eva/pipeline/io/VepProcessTest.java
```diff
@@ -80,26 +85,32 @@ private String getVariantInVepInputFormat(EnsemblVariant ensemblVariant) {
...
 
     @Test
     public void testWorkflowCloseWithoutOpening() throws Exception {
-        VepProcess vepAnnotationFileWriter = new VepProcess(annotationParameters, CHUNK_SIZE, VEP_TIMEOUT);
-        vepAnnotationFileWriter.close();
+        StringWriter writer = new StringWriter();
+        boolean skipComments = true;
+        VepProcess vepProcess = new VepProcess(annotationParameters, CHUNK_SIZE, VEP_TIMEOUT, writer, skipComments);
+        vepProcess.close();
     }
```
pabarcgar: Not in the scope of this PR: this class seems a critical piece in the annotation process, and it should have more tests

pabarcgar: Well, I see the class is tested not directly, but from VepAnnotationProcessorTest, right?

src/test/java/uk/ac/ebi/eva/pipeline/jobs/steps/processors/VepAnnotationProcessorTest.java
```diff
@@ -181,11 +139,11 @@ public void testVepTimeouts() throws Exception {
         annotationParameters.setVepPath(getResource("/mockvep_writeToFile_delayed.pl").getAbsolutePath());
 
         long vepTimeouts = 1;
-        VepAnnotationFileWriter vepAnnotationFileWriter = new VepAnnotationFileWriter(annotationParameters,
-                chunkSizeGreaterThanActualVariants, vepTimeouts);
+        VepAnnotationProcessor vepAnnotationProcessor = new VepAnnotationProcessor(annotationParameters,
+                                                                                   chunkSizeGreaterThanActualVariants, vepTimeouts);
```
pabarcgar: This line lenght is 133 (120 max in our standards)

jmmut:
I did more testing on longer files and can't reproduce the freezes, so let's move this on. I also added timestamps to be able to debug the issue about losing the mongo cursor.

PR #126, opened 2017-08-15
EVA-880 Reduce chance of timeouts in (re-)annotation search

cyenyxe:
When a big study is loaded and only a small amount of variants are pending to annotate, the search could timeout. In order to reduce the chance of this happening, an additional filter has been added to `VariantsMongoReader`. Now it restricts by file (being loaded) in addition to the study.

I'm not sure about the nesting (if no study is specified it doesn't matter whether the file is) as a concept, but in terms of performance is the only thing that makes sense...


jmmut:
I think this is OK, but wanted to highlight that from now on, the job for loading a new file with annotate.skip=false, will only annotate that file, not the whole study.

This is relevant to those cases where we skip annotation for all the files except the last one

PR #129, opened 2017-10-26
EVA-942 - Ignore Variant IDs submitted through VCFs

sundarvenkata-EBI:
VariantVcfFactory.java, VariantVCFFactoryTest.java - Ignore variant IDs submitted through VCFs (see EVA-942)


src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VariantVcfFactory.java
```diff
@@ -66,7 +66,9 @@
 
         String chromosome = getChromosomeWithoutPrefix(fields);
         int position = getPosition(fields);
-        Set<String> ids = getIds(fields);
+        //EVA-942 - Ignore IDs submitted through VCF
+        //Set<String> ids = getIds(fields);
```
sundarvenkata-EBI: Explicitly commented out (and did not delete) the previous line since what we are doing might be counter-intuitive to someone reading the code without history

PR #135, opened 2018-03-01
ignored all vcf records that reported non variant sites

Geekcoder555:
Description:
I have put condition checks to ignore all such vcf records for which conditions hold true.
fixes #132

PR #136, opened 2018-03-01
Ignored all vcf records which reported non variant sites

### LINKED ISSUE #132: Ignore VCF records that report non-variant sites
Since EVA is a variation archive, we shall not load into the database those records from a VCF file where one of the following conditions occurs:

* All the sample genotypes are non-variant (./., 0/0, 0|0)
* The value of the INFO field AF is zero
* The value of the INFO field AN is zero

AnkitKhatri55555:
Description:
Ignored all vcf records which reported non variant sites, by putting  some condition checks and ignoring those records for which condition holds true.
fixes #132 

file modified:
    https://github.com/EBIvariation/eva-pipeline/blob/develop/src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/VariantVcfFactory.java


cyenyxe:
Please push new commits to the branch in PR #135, do not create another PR for every code change.

Issue #143, opened 2018-11-08
Abort when duplicate sample names are found


cyenyxe:
Letting a VCF with duplicate samples be run through the pipeline causes issues when accessing them  through the API. The reason is that we use the sample name and not the column index they occupy in the file as an accessor; it is not possible for it to differentiate between 2 occurrences of the sample same.

In addition to this, the VCF specification doesn't allow samples with duplicate names to be listed in a file.

An example of a valid VCF header line would be:

> \#CHROM POS ID REF ALT QUAL FILTER INFO FORMAT SAMPLE1 SAMPLE2 SAMPLE3

And an invalid VCF header line would be:

> \#CHROM POS ID REF ALT QUAL FILTER INFO FORMAT SAMPLE1 SAMPLE2 SAMPLE1

The pipeline must abort when an input like the latter is provided, by modifying the class [VcfHeaderReader](https://github.com/EBIvariation/eva-pipeline/blob/develop/src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfHeaderReader.java#L103-L110).

PR #144, opened 2018-11-14
Abort when duplicate sample names are found #143

nitrozyna:
**Abort when duplicate sample names are found #143 solution comments**

I had a few issues compiling the project, in particular two test cases failing. Since I thought they weren't key to my implementation of the task, I commented them out.

Failed tests:
  CreateDatabaseIndexesStepTest.testIndexesAreCreated
  StatisticsMongoWriterTest.shouldCreateIndexesInCollection

I implemented the check for repeated sample names with a HashSet data structure. The reason why I used a Hashset is because HashSet is much faster than TreeSet (constant-time versus log-time for most operations) but offers no ordering guarantees which as far as I understand does not seem to be a requirement in the pipeline.

I couldn't find any suitable exceptions under src/main/java/uk/ac/ebi/eva/pipeline/runner/exceptions/ hence I decided to throw an IOException since that best describes the encountered problem.

I created a new folder to test on under the following name: same_sample_names.vcf.gz which contains a duplicate of sample: HG00096.

I also created a new test case in VcfHeaderReaderTest.java to test for duplicate samples.

Improvement thoughts:
If required, I could find all the duplicate samples, since for now the IOException is thrown on the first duplicate sample encountered.


src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfHeaderReader.java
```diff
@@ -98,19 +98,33 @@ public VariantSourceEntity read() throws Exception {
...
+            }
+            uniformSamples.add(sample);
+        }
         return new VariantSourceEntity(source);
+
     }
 
     @Override
     public void open(ExecutionContext executionContext) throws ItemStreamException {
+
```
cyenyxe: Remove blank lines at beginning and end of methods.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfHeaderReader.java
```diff
@@ -25,7 +25,7 @@
 import org.springframework.core.io.Resource;
 
 import uk.ac.ebi.eva.commons.models.data.VariantSourceEntity;
-
+import java.util.*;
```
cyenyxe: Use explicit imports following our internal coding guidelines. Can be configured in the IDE.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfHeaderReader.java
```diff
@@ -98,19 +98,33 @@ public VariantSourceEntity read() throws Exception {
...
+    private VariantSourceEntity doRead() throws IOException {
         if (variantReader == null) {
             throw new IllegalStateException("The method VcfHeaderReader.open() should be called before reading");
         }
         variantReader.pre();
         source.addMetadata(VARIANT_FILE_HEADER_KEY, variantReader.getHeader());
+
+        List<String> sampleNames = variantReader.getSampleNames();
+
+        Set<String> uniformSamples = new HashSet<String>();
```
cyenyxe: A shorter way would be to create the Set using the List as input, and comparing the sizes. To retrieve (possibly multiple) duplicate samples, you can use other methods in those classes.

I would also suggest renaming this variable to `uniqueSamples`, because that expresses the idea a bit better than `uniform`.

src/test/java/uk/ac/ebi/eva/pipeline/io/readers/VcfHeaderReaderTest.java
```diff
@@ -48,6 +51,20 @@
     @Rule
     public PipelineTemporaryFolderRule temporaryFolderRule = new PipelineTemporaryFolderRule();
 
+    @Test(expected = IOException.class)
+    public void testDuplicates() throws Exception {
```
cyenyxe: `testDuplicateSamples` to make clear that you are not referring to, for instance, duplicate records.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfHeaderReader.java
```diff
@@ -98,19 +98,33 @@ public VariantSourceEntity read() throws Exception {
...
+
+        for (String sample : sampleNames) {
+            if ((uniformSamples.contains(sample))) {
+                throw new IOException("Duplicate sample name given: " + sample);
```
cyenyxe: As you mention in the PR comments there was no suitable exception yet to cover this. But this one is too generic, because it could be thrown due to any issue with access to disk, etc. Maybe you could create a `DuplicateSamplesFoundException` inside this package (it wouldn't make sense to add it to `runner`)?

cyenyxe:
It seems the test that expects and exception is now failing because it is not thrown.

nitrozyna:
Sorry, I haven't finished yet. Once I do, I will submit a pull request again.

src/main/java/uk/ac/ebi/eva/pipeline/runner/exceptions/DuplicateSamplesFoundException.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2015-2017 EMBL - European Bioinformatics Institute
```
cyenyxe: Copyright 2018, because that is the year the file was created in.

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfHeaderReader.java
```diff
@@ -100,12 +104,22 @@ public VariantSourceEntity read() throws Exception {
...
         if (variantReader == null) {
             throw new IllegalStateException("The method VcfHeaderReader.open() should be called before reading");
         }
         variantReader.pre();
         source.addMetadata(VARIANT_FILE_HEADER_KEY, variantReader.getHeader());
+
+        List<String> sampleNames = new ArrayList<String> ( variantReader.getSampleNames());
+        HashSet<String> uniqueSampleNames = new HashSet<String>(sampleNames);
+
+        if (sampleNames.size() != uniqueSampleNames.size()){
```
cyenyxe: Missing white space before the `{`

src/test/java/uk/ac/ebi/eva/pipeline/io/readers/VcfHeaderReaderTest.java
```diff
@@ -48,6 +51,19 @@
...
 
+    @Test(expected = DuplicateSamplesFoundException.class)
+    public void testDuplicateSamples() throws Exception {
+        File input = getResource(INPUT_FILE_PATH_DUPLICATES);
+
+        VariantStudy.StudyType studyType = VariantStudy.StudyType.COLLECTION;
+        VariantSource.Aggregation aggregation = VariantSource.Aggregation.NONE;
+
+        VcfHeaderReader headerReader = new VcfHeaderReader(input, FILE_ID, STUDY_ID, STUDY_NAME,
+                studyType, aggregation);
```
cyenyxe: I noticed the method below does the same as this one, but others below use the constants. I think the code would look more clear if we use directly the constants; could you please also change the method right below?

nitrozyna: Could you elaborate? I am not sure if I understand what you mean here. I thought I am using constants the same way as other methods.

nitrozyna: Upon further revision I think what you requested is that I remove the indirection through the "studyType" and "aggregation" locally scoped variables for VariantStudy.StudyType.COLLECTION; and  VariantSource.Aggregation.NONE; respectively.

I initially took the indirect route as the fully qualified constant names are rather verbose and appear multiple times ( in the setup and final assertions ). Additionally these benefit from enhanced visibility of being set once in the test preamble. 

One compromise would be to adopt the same convention as that used for FILE_ID , STUDY_ID, etc. 

That is :  
Declare a set of private constants in the Test class and used these as indirections to the fully qualified constants. This has several benefits:
            - static final fields won't take up extra stack space per test 
            - the compiler will likely eliminate the intermediate storage anyway 
            - the developers can quickly asses and change the studyType and aggregation if required

Could you please confirm if you would prefer the private static field approach , or just the one that is established in the final test?

cyenyxe: Oops, you were right in using the indirections. I didn't notice the asserts below were using them.

nitrozyna: That's alright, thanks! I changed it back.

src/main/java/uk/ac/ebi/eva/pipeline/runner/exceptions/DuplicateSamplesFoundException.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package uk.ac.ebi.eva.pipeline.runner.exceptions;
+import java.util.List;
+
+/**
+ * Exception thrown where a string value argument is not recognized or valid in a function.
```
jmmut: Is this message correct? looks unrelated to me

src/main/java/uk/ac/ebi/eva/pipeline/io/readers/VcfHeaderReader.java
```diff
@@ -100,12 +104,22 @@ public VariantSourceEntity read() throws Exception {
...
         }
         variantReader.pre();
         source.addMetadata(VARIANT_FILE_HEADER_KEY, variantReader.getHeader());
+
+        List<String> sampleNames = new ArrayList<String> ( variantReader.getSampleNames());
+        HashSet<String> uniqueSampleNames = new HashSet<String>(sampleNames);
+
+        if (sampleNames.size() != uniqueSampleNames.size()) {
+            for (String str : uniqueSampleNames) {
+              sampleNames.remove(str);
```
sundarvenkata-EBI: While this is functional, the performance complexity of remove is O(n). Since this remove operation is inside a loop, complexity here is O(n*n). This can escalate quickly even for studies with moderately large number of samples (say 1000 genomes). @cyenyxe Any thoughts?

nitrozyna: Fair point! Since a HashSet is usually backed by a HashMap I could indeed just check for duplication while inserting into the HashSet rather than by removing from the slower ArrayList.
HashSet's add(obj) returns a status boolean showing whether the object was a duplicate or not. So I can just iteratively add the sample names rather than use a copy constructor and then check for any true flags returned at the end. Does that sound reasonable?

src/main/java/uk/ac/ebi/eva/pipeline/runner/exceptions/DuplicateSamplesFoundException.java
nitrozyna: Well spotted, thanks!

PR #150, opened 2019-03-24
Replace Annotation Metadata

108krohan:
- Iteratively success testing Travis CLI with 1 file change


src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/AnnotationLineMapper.java
```diff
@@ -62,29 +63,24 @@ public AnnotationLineMapper(String vepVersion, String vepCacheVersion) {
...
      *  - parseFrequencies is always true and the all line is always parsed;
      *  - The logic to move around the file (read line) and reference to previous line (currentVariantString) are removed;
      */
     @Override
-    public Annotation mapLine(String line, int lineNumber) {
-        ConsequenceType consequenceType = new ConsequenceType();
+    public AnnotationMongo mapLine(String line, int lineNumber) {
+        ConsequenceTypeMongo consequenceType = new ConsequenceTypeMongo(null, null,
+            null, null, null, null, null, null,
+            null, null, null, null, null, null);
```
sundarvenkata-EBI: Why not use the Empty constructor for ConsequenceTypeMongo?

pom.xml
```diff
@@ -113,6 +113,16 @@
                 </exclusion>
             </exclusions>
         </dependency>
```
sundarvenkata-EBI: Any reason why the <DependencyManagement> tag isn't included?

src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/AnnotationLineMapper.java
```diff
@@ -101,16 +97,38 @@ public Annotation mapLine(String line, int lineNumber) {
...
+                vepCacheVersion),
+            variantMap.get("chromosome"),
+            Integer.valueOf(variantMap.get("start")),
+            Integer.valueOf(variantMap.get("end")),
+            vepVersion,
+            vepCacheVersion,
+            null,
+            Collections.singleton(consequenceType));
+
+        AnnotationMongo AnnotationWithXref = new AnnotationMongo(
```
sundarvenkata-EBI: I am not sure what this additional annotation is about??

src/test/java/uk/ac/ebi/eva/commons/models/converters/data/VariantSourceEntryToDBObjectConverterTest.java
```diff
@@ -63,6 +63,7 @@ public void setUp() {
         file.addAttribute("QUAL", "0.01");
         file.addAttribute("AN", "2");
         file.addAttribute("MAX.PROC", "2");
+        file.setFormat("gzip");
```
sundarvenkata-EBI: Why this?

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/AnnotationMongoWriterTest.java
```diff
@@ -188,17 +188,27 @@ public void shouldWriteAllFieldsIntoMongoDbMultipleSetsAnnotations() throws Exce
...
+        AnnotationMongo annotation = new AnnotationMongo(
+            AnnotationMongo.buildAnnotationId("X", 1, "A", "T",
+                VEP_VERSION, VEP_CACHE_VERSION),
+            "X",
+            1,
+            10,
+            VEP_VERSION,
+            VEP_CACHE_VERSION,
+            null,
+            Collections.singleton(consequenceType));
```
sundarvenkata-EBI: Would moving this up, to where the Annotation instantiation originally was, have any impact?

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/AnnotationMongoWriterTest.java
```diff
@@ -188,17 +188,27 @@ public void shouldWriteAllFieldsIntoMongoDbMultipleSetsAnnotations() throws Exce
...
+
+        annotation = annotation.concatenate(annotation);
```
sundarvenkata-EBI: Again I am a bit unclear on why we append an Annotation object to itself...

src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/AnnotationLineMapper.java
108krohan: Use of an empty constructor for ConsequenceTypeMongo results in build error - (actual and formal argument lists differ in length). This is because the default constructor isn't public.

[PR - 81](https://github.com/EBIvariation/variation-commons/pull/81) tries to address this issue. 

The error-message is as follows:  
```constructor uk.ac.ebi.eva.commons.mongodb.entities.subdocuments.ConsequenceTypeMongo.ConsequenceTypeMongo(java.lang.String,java.lang.String,java.lang.String,java.lang.String,java.lang.String,java.lang.Integer,java.lang.Integer,java.lang.Integer,java.lang.String,java.lang.String,uk.ac.ebi.eva.commons.mongodb.entities.subdocuments.ScoreMongo,uk.ac.ebi.eva.commons.mongodb.entities.subdocuments.ScoreMongo,java.util.Set<java.lang.Integer>,java.lang.Integer) is not applicable
(actual and formal argument lists differ in length)
```

pom.xml
108krohan: I did not include the DependencyManagement tag because it results in test errors if I do so The test errors say they cannot find the configurations. This might possibly be because eva-pipeline's POM already has a parent POM - for reference: [Spring Boot Dependency Management with a Custom Parent](https://www.baeldung.com/spring-boot-dependency-management-custom-parent).

Also, Travis CLI build passes without the tag. So the other reason was to keep the eva-pipeline POM lean. 

To test if I could add dependencyManagement, I added the following to eva-pipeline's POM, which results in test errors (because it is unable to find Spring configurations). 
I'm new to Maven so I'm not 100% sure about my reasons. So please find the code used to include the tag:
```
    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>uk.ac.ebi.eva</groupId>
                <artifactId>variation-commons</artifactId>
                <version>0.7-SNAPSHOT</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>
```

Errors received were like so: 
```
java.lang.NoSuchMethodError: org.springframework.util.ClassUtils.getQualifiedMethodName(Ljava/lang/reflect/Method;Ljava/lang/Class;)Ljava/lang/String;
        at org.springframework.transaction.interceptor.TransactionAspectSupport.methodIdentification(TransactionAspectSupport.java:396)
```

src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/AnnotationLineMapper.java
108krohan: For reference: https://github.com/EBIvariation/variation-commons/blob/fda0d8bbf44f029ab7ef1f756f31964b3bc9e2c5/variation-commons-mongodb/src/main/java/uk/ac/ebi/eva/commons/mongodb/entities/subdocuments/ConsequenceTypeMongo.java#L99

src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/AnnotationLineMapper.java
108krohan: This additional annotation object is because of the way AnnotationMongo.java's class methods are defined in VariationCommons. The `addConsequenceType` and `addConsequenceTypes` functions are private. They are made accessible via `concatenate` command, which accesses `addConsequenceTypes`. `addConsequenceTypes` accesses private `addConsequenceType`.

Additional annotation object is created because as of now, we need to call `concatenate` to access `addConsequenceType` of `AnnotationMongo`. And `concatenate` accepts an `Annotation` as parameter. That is also why 1st I create `ConsequenceTypeMongo`, and then pass it into constructor of `Annotation` object. 


Why we need additional annotation object is because when `concatenate` method also updates `XrefMongo` attribute. That's why I named the additional object `AnnotationWithXref` and did not pass `currentAnnotation` directly.

Referring code snippet [AnnotationMongo.java#L207](https://github.com/EBIvariation/variation-commons/blob/fda0d8bbf44f029ab7ef1f756f31964b3bc9e2c5/variation-commons-mongodb/src/main/java/uk/ac/ebi/eva/commons/mongodb/entities/AnnotationMongo.java#L207)
```
    public AnnotationMongo concatenate(AnnotationMongo annotation) {
        AnnotationMongo temp = new AnnotationMongo(this);
        if (annotation.getConsequenceTypes() != null) {
            temp.addConsequenceTypes(annotation.getConsequenceTypes());
        }
        return temp;
    }
```

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/AnnotationMongoWriterTest.java
108krohan: Yes, when I test with moving this up to where the Annotation was originally instantiated, it is not possible for me to assign `consequenceType` object to the Annotation object. This is because AnnotationMongo does not have a setter, and the `addConsequenceType` method is private as of now (accessible only via `concatenate` method. But `concatenate` method also expects an `annotation` object which has it's `consequenceType` field populated. So the only way to set `consequenceType` is via AnnotationMongo's constructor. So that's why it's instantiation now happens after consequenceType. So consequence type can be passed as parameter in the constructor of AnnotationMongo.

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/AnnotationMongoWriterTest.java
108krohan: Concatenate method calls within it `addConsequenceTypes` which calls `addConsequenceType` for each consequenceType in the consequenceType set. `AddConsequenceType` also sets the Xref fields for AnnotationMongo object, besides just copying over ConsequenceType information. So that's why we call the `concatenate` function, and not return the annotation object directly (so as to set Xref fields as well).

src/test/java/uk/ac/ebi/eva/commons/models/converters/data/VariantSourceEntryToDBObjectConverterTest.java
108krohan: This is so because of the new VariantSourceEntryMongo. The `samp` field doesn't populate if `format` field of `file` isn't set. So that the `samp` field doesn't populate, which results in following test assertion-error: 

VariantSourceEntryToDBObjectConverterTest.testConvertToStorageTypeWithSamples:114 expected:<{ "fid" : "f1" , "sid" : "s1" , "attrs" : { "QUAL" : "0.01" , "AN" : "2" , "MAX?PROC" : "2"} , "samp" : { "def" : "0/0" , "0/1" : [ 1] , "1/1" : [ 2]}}> but was:<{ "fid" : "f1" , "sid" : "s1" , "attrs" : { "QUAL" : "0.01" , "AN" : "2" , "MAX?PROC" : "2"}}>

Added the following line to correct this: 
```file.setFormat("gzip");```

which then allowed for `samp` to be set, but now also required that `mongoFileWithIds` has `format` field present, as you can see in following test assertion-error message:
```
VariantSourceEntryToDBObjectConverterTest.testConvertToStorageTypeWithSamples:115 expected:<{ "fid" : "f1" , "sid" : "s1" , "attrs" : { "QUAL" : "0.01" , "AN" : "2" , "MAX?PROC" : "2"} , "samp" : { "def" : "0/0" , "0/1" : [ 1] , "1/1" : [ 2]} , **"fm" : "gzip"**}> but was:<{ "fid" : "f1" , "sid" : "s1" , "attrs" : { "QUAL" : "0.01" , "AN" : "2" , "MAX?PROC" : "2"}}>
```
So also added `mongoFileWithIds.put("fm", "gzip");` line to `mongoFileWithIds` because now we have `file.setFormat("gzip");`.

pom.xml
108krohan: Hi, please have a look at this comment thread. 

My current understanding (something I haven't tried yet) is to include dependencies via a singular variation-commons. 

So for example, only the following tag should suffice:
<dependency>
    <groupid>uk.ac.eba.eva</groupid>
    <artifactid>variation-commons</artifactid>
    <version>0.7-SNAPSHOT</version>
</dependency>

src/main/java/uk/ac/ebi/eva/pipeline/io/mappers/AnnotationLineMapper.java
108krohan: Hi Sundar, I was mulling over this comment and I found the following SO article useful: https://stackoverflow.com/questions/19395973/maven-adding-a-reference-to-a-parent-pom-project

The last two paragraphs in the answer are: 

"Update: You can do that in a limited way only for the dependencyManagement area with the scope import but this is only intended for using in the depdencyManagement and not for dependencies. An other solution might be to create a separate dummy project which has the dependencies too two childs as transitive deps. But this is not really beauty.

So the solution is simply to add the two or more dependencies directly which you are using in you project and that's it.

Parent POM isn't for that. To create a meta-project, create a regular Maven project with a name like common-libraries and reference all of your other projects as dependencies. Parent POM is to build several libraries in one step, and automatically use the correct build order."

What this suggests is to add the the dependencies directly. I just wanted to know how this can be resolved by tonight or tomorrow.

PR #155, opened 2019-11-17
EVA-1063 - Populate a default locus range (for use in Variant Browser) for a given species

sundarvenkata-EBI:


src/main/resources/utilities/populate_default_locus_range_for_species.py
```diff
@@ -0,0 +1,134 @@
...
+                "_id": 0
+            }
+        },
+        {
+            "$sort": {"num_entries": -1}
+        },
+        {
+            "$group": {
+                "_id": None,
+                "topChrStartPick": {"$first": "$$ROOT"}
```
jmmut: Are the 2 last stages making something different than `..., {"$sort": {"num_entries": -1}}, {"$limit:1}]`?

src/main/resources/utilities/populate_default_locus_range_for_species.py
```diff
@@ -0,0 +1,134 @@
...
+# limitations under the License.
+
+import click
+import configparser
+import urllib.parse
+
+from pymongo import MongoClient
+from __init__ import *
+
+DEFAULT_LOCUS_RANGE_COLLECTION_NAME = "defaultLocusRange"
```
jmmut: good idea. I assume you plan to make this available through some new webservice to un-hardcode it from the website?

src/main/resources/utilities/populate_default_locus_range_for_species.py
```diff
@@ -0,0 +1,134 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import click
+import configparser
+import urllib.parse
+
+from pymongo import MongoClient
```
jmmut: can you add a requirements.txt and readme for installing the dependencies?

src/main/resources/utilities/populate_default_locus_range_for_species.py
sundarvenkata-EBI: Yes. I intend to add a separate end point /default-locus-range similar to /annotation for a given species.

src/main/resources/utilities/populate_default_locus_range_for_species.py
sundarvenkata-EBI: I was not sure if $limit guarantees ordering or it retrieves just one entry at random so I opted for $first. But looks like limit() does honor the order. so I will change this....

jmmut: yep, also a great optimization will be done:

> When a $sort precedes a $limit and there are no intervening stages that modify the number of documents, the optimizer can coalesce the $limit into the $sort. This allows the $sort operation to only maintain the top n results as it progresses, where n is the specified limit, and ensures that MongoDB only needs to store n items in memory. This optimization still applies when allowDiskUse is true and the n items exceed the aggregation memory limit.

PR #163, opened 2021-02-04
EVA-2287 Avoid removing chr prefix

andresfsilva:
Unfortunately `eva-pipeline` is not using the `variation-commons` library and have its own implementation for the Readers although they seem very similar. We should probably refactor this repo and use variation commons. See [EVA-2300](https://www.ebi.ac.uk/panda/jira/browse/EVA-2300)


tcezard (approved):

sundarvenkata-EBI (approved):

sundarvenkata-EBI:
Sorry just noticed this PR now!

PR #180, opened 2022-01-11
EVA-2581 - only include id and shard key in SimplifiedAnnotation for safe upserts

apriltuesday:


tcezard (commented):
This looks good, but I'm wondering if the bug could be reproduced in a test?

apriltuesday:
As we suspected, I haven't been able to replicate the bug in the test setting... if anyone has suggestions I'm happy to implement them.

nitin-ebi (approved):

nitin-ebi (approved):

tcezard (approved):

sundarvenkata-EBI (approved):

andresfsilva (approved):

PR #181, opened 2022-01-28
EVA-2729: upsert into files collection

apriltuesday:


andresfsilva (approved):

sundarvenkata-EBI (approved):
Looks good! However, I feel that we should trigger updates only when a specific parameter (say for ex: forceReprocessing) is passed. We can do that in another PR though.

tcezard:
> Looks good! However, I feel that we should trigger updates only when a specific parameter (say for ex: forceReprocessing) is passed. We can do that in another PR though.

I'm curious as to why. Is there a risk or associated with upserting?

sundarvenkata-EBI:
> I'm curious as to why. Is there a risk or associated with upserting?

As @apriltuesday had pointed out before, there is an implicit assumption that the file being processed has not changed between multiple executions. But that might not be the case every time. The cost of recovering from accidental mixed processing of old and new files is pretty high (particularly for large submissions). So we can make an argument to make our intention explicit and trigger upserts only when that argument is passed in.

PR #188, opened 2023-01-19
EVA-3143: eva-pipeline gitlab project for CI/CD

nitin-ebi:
There are some test cases (10 to be exact) which needed to be turned off. These test cases were setting and then testing for file permission like not readable and not writable. These are failing in gitlab because our gitlab-runner and build run as root and has extra permissions allowing it to read and write file which are set to be not readable and not writable.


sundarvenkata-EBI (approved):

.gitlab-ci.yml
```diff
@@ -10,11 +10,22 @@ variables:
     URL_MAVEN_SETTINGS: https://api.github.com/repos/EBIvariation/configuration/contents/eva-maven-settings.xml
     MEDIA_TYPE: application/vnd.github.raw
 
+
 test:
     stage: test
-    image: maven:3.6.1-jdk-8-alpine
+    image: ubuntu:20.04
```
sundarvenkata-EBI, line 16: I think we might have to standardize on one OS... If that happens to be Ubuntu, then we might want to change eva-accession's CI/CD scripts to the same OS...

nitin-ebi: Could not use maven:3.6.1-jdk-8-alpine here, because some of the test cases which were reading binary files from test resources where failing. Didn't face any such issue when using ubuntu.

Completely agree with your point of using a standard image across our builds. My vote is for ubuntu because that is closer to our local environment in which we build and test our applications.

sundarvenkata-EBI (approved):

tcezard (approved):

PR #192, opened 2023-10-16
EVA-3425 make sure reference and alternate is always upper case

nitin-ebi:


src/main/java/uk/ac/ebi/eva/commons/models/data/Variant.java
```diff
@@ -135,6 +135,8 @@ public Variant(String chromosome, int start, int end, String reference, String a
         this.chromosome = chromosome;
         this.start = start;
         this.end = end;
+        reference = Objects.nonNull(reference) ? reference.toUpperCase() :  null;
```
sundarvenkata-EBI, line 138: Guessing this will not mutate the arguments that are passed since String is a primitive type. Could you double check?

src/main/java/uk/ac/ebi/eva/commons/models/data/VariantSourceEntry.java
```diff
@@ -82,7 +83,10 @@ public VariantSourceEntry(String fileId, String studyId) {
     public VariantSourceEntry(String fileId, String studyId, String[] secondaryAlternates, String format) {
         this.fileId = fileId;
         this.studyId = studyId;
-        this.secondaryAlternates = secondaryAlternates;
+        this.secondaryAlternates = Arrays.stream(secondaryAlternates)
```
sundarvenkata-EBI, line 86: We should call setSecondaryAlternates here...

src/main/java/uk/ac/ebi/eva/commons/models/data/VariantSourceEntry.java
```diff
@@ -111,7 +115,10 @@ public String[] getSecondaryAlternates() {
     }
 
     public void setSecondaryAlternates(String[] secondaryAlternates) {
-        this.secondaryAlternates = secondaryAlternates;
+        this.secondaryAlternates = Arrays.stream(secondaryAlternates)
+                .map(a->a.toUpperCase())
+                .collect(Collectors.toList())
+                .toArray(new String[0]);;
```
sundarvenkata-EBI, line 121:
```suggestion
                .toArray(new String[0]);
```

sundarvenkata-EBI (commented):
Looks good but we might need to write explicit tests given that we are modifying something very fundamental...

tcezard (approved):

src/main/java/uk/ac/ebi/eva/commons/models/data/VariantSourceEntry.java
nitin-ebi: done

src/main/java/uk/ac/ebi/eva/commons/models/data/VariantSourceEntry.java
nitin-ebi: corrected

src/main/java/uk/ac/ebi/eva/commons/models/data/Variant.java
nitin-ebi: String being an immutable class, it will not mutate the existing object but will create a new one when **"toUpperCase()"** is called

sundarvenkata-EBI (approved):

PR #220, opened 2026-04-17
EVA-4137 Allow study name to have non printable characters

nitin-ebi:


src/main/java/uk/ac/ebi/eva/pipeline/parameters/InputParameters.java
```diff
@@ -83,7 +83,8 @@ public String getVcfId() {
     }
 
     public String getStudyName() {
-        return studyName;
+        // Strip anything non-ASCII printable except newline and carriage return
+        return studyName == null ? null : studyName.replaceAll("[^\\x20-\\x7E\\n\\r]", "");
```
nitin-ebi, line 87: We need to decide on how strict we want to be with the regex

src/main/java/uk/ac/ebi/eva/pipeline/parameters/validation/ParametersValidatorUtil.java
```diff
@@ -31,23 +31,9 @@ public class ParametersValidatorUtil {
...
                                    String jobParametersName) throws JobParametersInvalidException {
         checkIsNotNullString(stringToValidate, jobParametersName);
-        checkDoesNotContainPrintableCharacters(stringToValidate, jobParametersName);
         checkLength(stringToValidate, jobParametersName);
     }
 
-    /**
-     * \n or \r are valid non-printable characters
-     */
-    static void checkDoesNotContainPrintableCharacters(String stringToValidate,
```
apriltuesday, line 41: I didn't realise we were actually checking for these characters ourselves - did you confirm via integration tests whether MongoDB will load them?

nitin-ebi: Haven't tested with integration tests because the tests in the Eva-Pipeline itself use MongoDB, and they can load these names without any issues.

apriltuesday: Makes sense. I was just wondering why we were doing these checks in the first place, but given that it's from [2017](https://github.com/EBIvariation/eva-pipeline/pull/86) and quite a few things have changed since then, it's probably not worth worrying too much about...

tcezard (approved):

apriltuesday (approved):

PR #225, opened 2026-05-15
EVA-4112 Upgrade project - java 21 - spring 3.4

nitin-ebi:


src/main/java/uk/ac/ebi/eva/pipeline/runner/EvaPipelineJobLauncherCommandLineRunner.java
```diff
@@ -197,20 +196,20 @@ private JobParameters getJobParameters(Properties commandLineProperties, Propert
...
         }
     }
 
     private Optional<Properties> getJobParametersFromCommandLine(String[] args)
             throws NotValidParameterFormatException {
-        checkAllParametersStartByDoubleDash(args);
-        String[] processedArgs = removeStartingHypens(args);
+        String[] processedArgs = Arrays.stream(args)
+                .map(arg -> arg.startsWith("--") ? arg.substring(2) : arg)
+                .toArray(String[]::new);
```
tcezard, line 211: I guess we don't care too much about enforcing that all parameters starts with `--`

pom.xml
```diff
@@ -167,9 +154,9 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
+                <version>3.2.5</version>
                 <configuration>
-                    <!-- Travis build workaround -->
-                    <argLine>-Xms1024m -Xmx2048m</argLine>
+                    <reuseForks>false</reuseForks>
```
tcezard, line 159: This will prevent the test from running in parallel. Is this required ?

tcezard (approved):

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/StatisticsMongoWriterTest.java
```diff
@@ -58,37 +57,53 @@
...
-    @Autowired
-    private MongoMappingContext mongoMappingContext;
+    private MongoTemplate mongoTemplate;
+
+    @BeforeEach
+    public void setUp() throws Exception {
+        mongoTemplate.getDb().drop();
+
+        // mongodb no longer creates index by default, it needs to created explicitly or enabled to be created
+        // without the compound index mongodb will consider these as 2 separate documents and insert both
```
apriltuesday, line 75: Do there need to be changes to the statistics writer itself because of this? Since we are creating a new database and statistics collection for each species/assembly pair.

apriltuesday (approved):

src/main/java/uk/ac/ebi/eva/pipeline/runner/EvaPipelineJobLauncherCommandLineRunner.java
nitin-ebi: Spring does not mandate -- for parameters. We should be able to pass the parameters with or without --

pom.xml
nitin-ebi: Putting **reuseForks** as false forces the mvn to spin up a new JVM for each test. It creates a separate isolated context for everything, so there is no cross-contamination between tests. 
Although it might make the tests a little slow, as there is extra overhead of creating a new JVM for each test.

src/test/java/uk/ac/ebi/eva/pipeline/io/writers/StatisticsMongoWriterTest.java
nitin-ebi: Leaving it as it is. 
I checked on this class, and it doesn't appear to be used anywhere. So we should remove it along with **StatisticsMongoWriter** and **PopulationStatistics**.

